PR #2, opened 2024-06-24
feat: Add Chainlit

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-261

## Changes
 - Add Chainlit as a dependency
 - Replace the Flask app with a basic Hello, world! Chainlit app
 - Update the Dockerfile and docker-compose files to run the Chainlit app and expose it on the correct port
 - Delete a number of test files that are now no longer necessary (because they tested the Flask app), as well as the OpenAPI CI step (because it dependent on the Flask app)

## Context for reviewers
 - The diff is not as big as it looks! The vast majority of those changes are either automatic changes to poetry.lock, or they are deleted test files.

## Testing
`make init start`, then at http://localhost:8000
image


ccheng26 (approved):
lgtm, tested and running successfully

PR #5, opened 2024-06-24
feat: Add MockSentenceTransformer

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-265

## Changes
 - Add SentenceTransformer package (the poetry.lock change is the vast majority of the diff)
 - Add configuration for the embedding model, plus instructions for how to cache it locally
 - Add MockSentenceTransformer class for use in tests + test coverage
 - Disabled the trivy scan -- it was failing because it was having trouble building the image.

## Context for reviewers
 - MockSentenceTransformer calculates the average word length of the input text, then converts this number to a range from 0 to 1, then converts that to an angle in two dimensions and uses that as the first two values of the embedding. This means that text with similar average word lengths will produce vectors with closer angles (direction). We then normalize the vector (make its magnitude 1) which lets us use dot product to calculate the similarity in direction between two vectors produced this way.

## Testing
There is a unit test for MockSentenceTransformer. Testing that `multi-qa-mpnet-base-dot-v1` is caching correctly is a bit trickier because we explicitly don't want to download this model for unit tests:
1. Following the set up instructions, from within the `app` directory, run `mkdir models && git clone https://huggingface.co/sentence-transformers/multi-qa-mpnet-base-dot-v1 models/multi-qa-mpnet-base-dot-v1`, then `make init start`
2. Run `make login`, which will put you in a shell in the running container, then `poetry run python` to open a Python console
3. This Python code should then run successfully:

```python
from sentence_transformers import SentenceTransformer
from src.app_config import AppConfig
assert AppConfig().embedding_model == "/app/models/multi-qa-mpnet-base-dot-v1"
SentenceTransformer(AppConfig().embedding_model)
```


ccheng26 (approved):
lgtm, tests passing and setup instructions run successfully

PR #9, opened 2024-06-25
feat: Add Document and Chunk models

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-264

## Changes
 - Add Document and Chunk models + database migration
 - Add DocumentFactory and ChunkFactory + tests
 - Minor documentation updates and small code tweaks (left comments inline on the PR)

## Testing
Added unit tests in `test_factories.py`


.gitignore
```diff
@@ -18,3 +18,6 @@
 
 # Python testing stuff
 *__pycache__*
+
+# Override that enables app/.env to be read for local development
+docker-compose.override.yml
```
KevinJBoyer, line 23: I think I accidentally dropped this line in an earlier PR while merging main 😱

app/src/db/models/document.py
```diff
@@ -0,0 +1,33 @@
...
+        "Chunk", back_populates="document", cascade="all, delete"
+    )
+
+
+class Chunk(Base, IdMixin, TimestampMixin):
+    __tablename__ = "chunk"
+
+    content: Mapped[str]
+    tokens: Mapped[int | None]
+    embedding: Mapped[np.ndarray] = mapped_column(Vector(768))
```
ccheng26, line 30: would this cause issues if we were to use a different embedding function?

KevinJBoyer: Good spot -- I think there are a couple of options:
 - One is to leave off the length and make it just `mapped_column(Vector)`. Pro of this is we can then insert values with other lengths, but the con is we'll have to make sure we're managing which rows came from which embedding fucntions
  - Another is to rename this column to something more specific, like `mpnet_embedding`, and if we want to switch to different embedding models in the future we add another column
  
  I'm open to both options -- what do you think?

ccheng26: I'm partial to option 2. The more I think of it, once we optimize we probably won't be needing to manage multiple embedding functions.

KevinJBoyer: 👍  just made this change!

ccheng26 (approved):
lgtm, tests run successfully

PR #13, opened 2024-06-28
feat: Ingest Guru Cards

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-269

## Changes
 - Add script and test coverage for ingesting Guru cards in .json format
 - Add documentation for how to run the script

## Context for reviewers
 - I'm not splitting the cards if they exceed the max sequence length at this time because we probably want to investigate what splitting strategy makes sense, and of the 429 Guru cards in our latest set, only 8 have more than 512 tokens (and the max is 681, so not much context is being lost for these 8.)

## Testing
In addition to test coverage, after running locally in my DB:

image

image

I also manually deployed this branch to the dev environment, then ran it and manually inspected the logs to confirm a successful run: https://us-east-1.console.aws.amazon.com/ecs/v2/clusters/app-dev/tasks/87b127b0a05142b2a09e57951cb1f292/logs?region=us-east-1


ccheng26 (approved):
lgtm, ingestion code run successfully

PR #15, opened 2024-06-28
feat: Augment response with retrieved cards

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-272

## Changes
 - Add retrieve() and test coverage
 - Add context keyword to generate() and test coverage
 - Integrate these into chainlit.py
 - Simplify the math behind the MockSentenceTransformer, and fix a bug in the ChunkFactory

## Testing
Spin up the app at localhost:8000/chat, then run `make logs` (after you've ingested Guru cards) and ask a question. The logs should show the queried cards:
image
image

### DIFF
--- a/app/tests/mock/test_mock_sentence_transformer.py
+++ b/app/tests/mock/test_mock_sentence_transformer.py
@@ -7,8 +7,6 @@ def test_mock_sentence_transformer():
     assert embedding_model.max_seq_length == 512
     assert embedding_model.tokenizer.tokenize("Hello, world!") == ["Hello,", "world!"]
     assert len(embedding_model.encode("Hello, world!")) == 768
-    # It should be about 1, but with some tolerance for floating point imprecision
-    assert sum(embedding_model.encode("Hello, world!")) - 1 < 0.01
 
     # Test that we can compare similarity with dot product,
     # where sentences with the same average length word are considered more similar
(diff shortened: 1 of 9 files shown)


app/tests/mock/test_mock_sentence_transformer.py
```diff
@@ -7,8 +7,6 @@ def test_mock_sentence_transformer():
     assert embedding_model.max_seq_length == 512
     assert embedding_model.tokenizer.tokenize("Hello, world!") == ["Hello,", "world!"]
     assert len(embedding_model.encode("Hello, world!")) == 768
-    # It should be about 1, but with some tolerance for floating point imprecision
-    assert sum(embedding_model.encode("Hello, world!")) - 1 < 0.01
```
KevinJBoyer, line 11: The model we're using (mpnet) isn't actually normalized, so I removed this test requirement

app/tests/src/test_generate.py
```diff
@@ -1,8 +1,10 @@
-from src.generate import generate, get_models
+from src.generate import PROMPT, generate, get_models
 from tests.mock import mock_completion
+from tests.src.db.models.factories import ChunkFactory
 
 
 def test_get_models(monkeypatch):
+    monkeypatch.delenv("OPENAI_API_KEY")
     assert get_models() == {}
```
KevinJBoyer, line 8: I noticed this test failed if I had a real API key defined in my `.env`

ccheng26 (approved):
lgtm!

PR #16, opened 2024-06-30
feat: Show accordions with cards

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-282

## Changes
 - Commit the `.chainlit` folder (this is the majority of the additions in this PR)
   - Enable the unsafe_html in the config, and set custom HTML and JS for accordions from USWDS. Also enable light mode and disable file uploads
 - Add a `format_guru_cards` function and append the output to the response of the user message
 - Refactor `retrieve` into two functions, `retrieve` and `retrieve_with_scores`. The initial implementation was `retrieve_with_scores`, but we don't use it yet, so this simplifies the code. Also moved logging out of chainlit.py and directly into these functions. Add test coverage for these.

## Testing
Tests added for formatter, manual testing screenshot below:
image

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -5,6 +5,7 @@ from sentence_transformers import SentenceTransformer
 import chainlit as cl
 import src.adapters.db as db
 from src.app_config import AppConfig
+from src.format import format_guru_cards
 from src.generate import generate
 from src.retrieve import retrieve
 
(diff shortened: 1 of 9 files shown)


app/src/chainlit.py
```diff
@@ -15,17 +16,18 @@
 
 @cl.on_message
 async def main(message: cl.Message) -> None:
-    logger.info(f"Received: {message.content}")
+    logger.info(f"Received: {message.content!r}")
```
yoomlam, line 19: TIL: https://stackoverflow.com/questions/44800801/in-python-format-f-string-strings-what-does-r-mean

app/src/format.py
```diff
@@ -0,0 +1,31 @@
...
+            aria-controls="a-{_accordion_id}"
+            >
+            <a href='https://link/to/guru_card'>{chunk.document.name}</a>
+        </button>
+    </h4>
+    <div id="a-{_accordion_id}" class="usa-accordion__content usa-prose" hidden>
+        <p>Summary: {chunk.document.content}</p>
+    </div>
+</div>"""
+    return "<h4>Related Guru cards</h4>" + cards_html
```
yoomlam, line 31: Odd semantic of `h4` nested under `h4`. Should this be:
```suggestion
    return "<h3>Related Guru cards</h3>" + cards_html
```

app/src/retrieve.py
```diff
@@ -6,14 +7,40 @@
...
+    logger.info(f"Retrieving context for {query!r}")
+
+    query_embedding = embedding_model.encode(query, show_progress_bar=False)
+
+    chunks = db_session.scalars(
+        select(Chunk).order_by(Chunk.mpnet_embedding.max_inner_product(query_embedding)).limit(k)
+    ).all()
+
+    for chunk in chunks:
+        logger.info(f"Retrieved: {chunk.document.name!r}")
```
yoomlam, line 25: Should this be `debug`?
```suggestion
        logger.debug(f"Retrieved: {chunk.document.name!r}")
```

yoomlam (approved):
Yay! Tests!

PR #20, opened 2024-07-08
feat: add ollama call to get_models

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-268

## Changes
- Added ollama call to get_models

## Testing

> run make test in `app/` directory


README.md
```diff
@@ -1,9 +1,23 @@
...
 
 ## Local Development
+
 To set up your local development environment, follow the instructions in [Getting Started](docs/app/getting-started.md).
 
+## Installing Ollama
+
+You can install ollama on macOSX by calling `brew install ollama`, for other operating systems or the desktop app see the [Ollama repository](https://github.com/ollama/ollama).
+
+To download a model run `ollama pull model_name`
```
yoomlam, line 13:
```suggestion
To download a model run `ollama pull <model_name>`
```

README.md
```diff
@@ -1,9 +1,23 @@
...
+
+Ex: `ollama pull llama2:7b`
+
+To start Ollama without the desktop app run `ollama serve`
+
+To configure a local secret to enable Ollama add `OLLAMA_HOST=http://host.docker.internal:11434` to your `.env` file.
```
yoomlam, line 19: Setting `OLLAMA_HOST` is not needed if running locally without Docker. Using `host.docker.internal` assumes you're running Docker, right?

KevinJBoyer: (I don't feel strongly about it, if folks do like having the app running locally, we can revise the approach in this ticket -- would need a way to communicate to `get_models` that Ollama is available as a local service, but that seems reasonable)

yoomlam: > would need a way to communicate to get_models that Ollama is available as a local service

Using the default `http://localhost:11434` should work.

yoomlam: @ccheng26 Reverting back to use `host.docker.internal` sounds like the simplest solution and aligns with the original intent in the ticket. [Related Slack thread](https://nava.slack.com/archives/C06DP498D1D/p1720473007216669?thread_ts=1720472024.809749&cid=C06DP498D1D)

app/src/generate.py
```diff
@@ -20,6 +21,12 @@ def get_models() -> dict[str, str]:
 
     if "OPENAI_API_KEY" in os.environ:
         return {"OpenAI GPT-4o": "gpt-4o"}
+    if "OLLAMA_HOST" in os.environ:
+        ollama_models = {}
+        models = ollama.list()
+        for model in models["models"]:
+            ollama_models[f"ollama/{model['name']}"] = model["name"]
```
KevinJBoyer, line 28: The key should be the human-readable name and the value should be what LiteLLM expects, so I think this should be:
```suggestion
            ollama_models[f"Ollama {model['name']}"] = f"ollama/{model['name']}"
```

README.md
```diff
@@ -1,9 +1,23 @@
 # Nava Labs Decision Support Tool
+
 Welcome! You are at the root of the Nava Labs Decision Support Tool pilot repo.
 
 ## Local Development
+
 To set up your local development environment, follow the instructions in [Getting Started](docs/app/getting-started.md).
 
+## Installing Ollama
```
KevinJBoyer, line 9:
```suggestion
### Installing Ollama
```
(since this is part of setting up Ollama)

KevinJBoyer (approved):
LGTM, thank you!

PR #22, opened 2024-07-11
feat: productionize uvicorn

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-281

## Changes

> modify configuration to use gunicorn instead of uvicorn

## Testing
Run `make start` then `docker ps` 
`labs-decision-support-tool-main-app   "poetry run gunicorn…"  0.0.0.0:8000->8000/tcp   main-app` should appear

or run `poetry  run  gunicorn -k uvicorn.workers.UvicornWorker -b 0.0.0.0:8000  src.app:app` directly from `/app` directory

App should be accessible at localhost:8000/chat


app/gunicorn.conf.py
```diff
@@ -0,0 +1,21 @@
...
+
+For more information, see https://docs.gunicorn.org/en/stable/configure.html
+"""
+
+from src.app_config import AppConfig
+
+app_config = AppConfig()
+
+bind = app_config.host + ':' + str(app_config.port)
+# Calculates the number of usable cores and doubles it. Recommended number of workers per core is two.
```
yoomlam, line 16: Where is the number of workers set?
https://github.com/navapbc/template-application-flask/blob/main/app/gunicorn.conf.py and https://docs.gunicorn.org/en/stable/configure.html show setting the workers explicitly.

app/gunicorn.conf.py
```diff
@@ -0,0 +1,21 @@
...
+    threads(int): The number of threads per worker for handling requests.
+
+For more information, see https://docs.gunicorn.org/en/stable/configure.html
+"""
+
+from src.app_config import AppConfig
+
+app_config = AppConfig()
+
+bind = app_config.host + ':' + str(app_config.port)
```
yoomlam, line 15: Since we're using the `-b 0.0.0.0:8000` argument when running in the Docker environment, is this `bind` variable only used when not using Docker?
Another related question is: does `host` and `port` need to be part of `app_config` if that configuration is always overridden by the `-b 0.0.0.0:8000` argument?

app/src/app_config.py
```diff
@@ -16,3 +16,5 @@ class AppConfig(PydanticBaseEnvConfig):
 
     embedding_model: str = "multi-qa-mpnet-base-dot-v1"
     global_password: str | None = None
+    host: str = "127.0.0.1"
```
yoomlam, line 19: FYI: I believe this `127.0.0.1` will only allow `localhost` connections.

app/gunicorn.conf.py
ccheng26: By [default](https://docs.gunicorn.org/en/stable/settings.html#workers) the number of worker values is 1. When I set the workers to 2+. I ended up with this error:
```
/Users/charlie/projects/dst/labs-decision-support-tool/app/.venv/lib/python3.12/site-packages/threadpoolctl.py:1214: RuntimeWarning: 
Found Intel OpenMP ('libiomp') and LLVM OpenMP ('libomp') loaded at
the same time. Both libraries are known to be incompatible and this
can cause random crashes or deadlocks on Linux when loaded in the
same Python program.
Using threadpoolctl may cause crashes or deadlocks. For more
information and possible workarounds, please see
    https://github.com/joblib/threadpoolctl/blob/master/multiple_openmp.md

  warnings.warn(msg, RuntimeWarning)
/Users/charlie/projects/dst/labs-decision-support-tool/app/.venv/lib/python3.12/site-packages/threadpoolctl.py:1214: RuntimeWarning: 
Found Intel OpenMP ('libiomp') and LLVM OpenMP ('libomp') loaded at
the same time. Both libraries are known to be incompatible and this
can cause random crashes or deadlocks on Linux when loaded in the
same Python program.
Using threadpoolctl may cause crashes or deadlocks. For more
information and possible workarounds, please see
    https://github.com/joblib/threadpoolctl/blob/master/multiple_openmp.md
```
I followed the work arounds in the site but those didn't seem to fix the issue for me, using the default 1 worker did.

app/gunicorn.conf.py
ccheng26: The variable should be used when we're not using Docker. The host and port can change based on the env params.

app/gunicorn.conf.py
KevinJBoyer: Setting workers = 4 works for me. Do you also get that error when you run it in Docker?

ccheng26: Running it on Docker seems to work, I was testing with the command `poetry  run  gunicorn -w 4 -k uvicorn.workers.UvicornWorker -b 0.0.0.0:8000  src.app:app`

app/gunicorn.conf.py
```diff
@@ -0,0 +1,22 @@
...
```
yoomlam, line 15:
```suggestion
# Since the -b 0.0.0.0:8000` argument is used when running in the Docker environment,
# this bind variable is only used when not using Docker
bind = app_config.host + ':' + str(app_config.port)
```

app/gunicorn.conf.py
```diff
@@ -0,0 +1,22 @@
...
+# Calculates the number of usable cores and doubles it. Recommended number of workers per core is two.
+# https://docs.gunicorn.org/en/latest/design.html#how-many-workers
+# We use 'os.sched_getaffinity(pid)' not 'os.cpu_count()' because it returns only allowable CPUs.
+# os.sched_getaffinity(pid): Return the set of CPUs the process with PID pid is restricted to.
+# os.cpu_count(): Return the number of CPUs in the system.
+threads = 4
+workers = 4
```
yoomlam, line 24: What's the reason to use `4`?
https://github.com/navapbc/template-application-flask/blob/dded29ce1ae0d3b9cd067a971850a290da29e0c2/app/gunicorn.conf.py#L24 calculates it.

ccheng26: I had an issue where os didn't have `'sched_getaffinity'` defined- this might have been another limitation of the system I was using. It seems to run without issue on the Docker container, so I'll add it here.

app/gunicorn.conf.py
```diff
@@ -0,0 +1,25 @@
...
+
+# Since the `-b 0.0.0.0:8000` argument is used when running in the Docker environment,
+# this bind variable is only used when not using Docker
+bind = app_config.host + ':' + str(app_config.port)
+# Calculates the number of usable cores and doubles it. Recommended number of workers per core is two.
+# https://docs.gunicorn.org/en/latest/design.html#how-many-workers
+# We use 'os.sched_getaffinity(pid)' not 'os.cpu_count()' because it returns only allowable CPUs.
+# os.sched_getaffinity(pid): Return the set of CPUs the process with PID pid is restricted to.
+# os.cpu_count(): Return the number of CPUs in the system.
+workers = len(os.sched_getaffinity(0)) * 2
```
yoomlam, line 24: How do I confirm there are N workers running?

yoomlam: Confirmed!

yoomlam (approved):

PR #23, opened 2024-07-15
fix: Conditionally import ollama

yoomlam:
## Ticket

If the ollama package doesn't exist, then don't cause `ModuleNotFoundError: No module named 'ollama'`.

## Changes

Conditionally import ollama, i.e., only when `OLLAMA_HOST` env variable is set.

## Testing

Dev environment shouldn't change since `ollama` was moved to dev dependencies.

Setting `OLLAMA_HOST` for prod will result in `ModuleNotFoundError`.


KevinJBoyer:
Seeing this branch and wanted to pop by to say Ollama is part of pyproject.toml and poetry.lock, so it should always exist. If you're seeing this error, you may need to run `make build` (which is required every time a new dependency is added.)

app/tests/src/test_generate.py
```diff
@@ -89,8 +21,9 @@ def test_get_models_ollama(monkeypatch):
     if "OPENAI_API_KEY" in os.environ:
         monkeypatch.delenv("OPENAI_API_KEY")
     monkeypatch.setenv("OLLAMA_HOST", "mock_key")
-    monkeypatch.setattr(ollama, "list", ollama_model_list)
 
+    sys.modules["ollama"] = __import__("mock_ollama")
+    # TODO: use `monkeypatch_module` so that sys.modules doesn't mess up other tests
```
yoomlam, line 26: @KevinJBoyer @ccheng26 Do y'all know how to use https://github.com/navapbc/labs-decision-support-tool/blob/5a46fd63f2788991f075e4a5088ffec4df148450/app/tests/conftest.py#L62

KevinJBoyer: I'm generally familiar, but did you see my comment [here](https://github.com/navapbc/labs-decision-support-tool/pull/23#issuecomment-2228899367)?

yoomlam:
> Seeing this branch and wanted to pop by to say Ollama is part of pyproject.toml and poetry.lock, so it should always exist. If you're seeing this error, you may need to run `make build` (which is required every time a new dependency is added.)

Since it doesn't make since to run Ollama in the production deployment, we should move the ollama package for dev dependencies, right?

app/tests/src/test_generate.py
KevinJBoyer: If you're moving ollama to [tool.poetry.group.dev.dependencies], do you need to change anything about this test? It looks like moto is part of the same group, and is imported in tests

KevinJBoyer:
> Since it doesn't make since to run Ollama in the production deployment, we should move the ollama package for dev dependencies, right?

Ah that makes sense. Looks like `python-dotenv` could also be moved into the dev dependencies group too.

yoomlam:
> > Since it doesn't make since to run Ollama in the production deployment, we should move the ollama package for dev dependencies, right?
> 
> Ah that makes sense. Looks like `python-dotenv` could also be moved into the dev dependencies group too.

I'll leave moving `python-dotenv` for a different PR since it could make since to use `.env` files in prod.

app/tests/src/test_generate.py
yoomlam: > do you need to change anything about this test?

I don't think so. I just need to address the `TODO` I added.

KevinJBoyer: Played around with this -- if you're comfortable with `import ollama` in `test_generate.py`, this test will continue to work as originally written. However, you can remove the import by mocking the function using a string for the path, like: `monkeypatch.setattr("ollama.list", ollama_list)`.

Unfortunately, mocking the module is tricky because this PR modifies the code in `generate.py` to import the module within a function, so there's no way to do a top-level mock which you would normally do via `monkeypatch.setattr("src.generate.ollama", tests.mock.mock_ollama)`

yoomlam: Reverted test files.

PR #25, opened 2024-07-15
feat: Use URL query 'engine' parameter to set chatbot's configuration

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-327

## Changes

- Add `ChatEngineInterface` abstract class that acts as an interface for chat engines
- Refactor existing chatbot backend to `GuruMultiprogramEngine` class
- Add placeholder engines: `GuruSnapEngine` and `PolicyMichiganEngine`

## Testing
* Go to `chat/` or `/chat/?engine=unknownEngine`, expect "Available engines: ['guru-snap', 'policy-mi', 'guru-multiprogram']"
* Go to `/chat/?engine=guru-multiprogram`, expect "Chat engine started: Guru Multi-program Chat Engine"
   - Expect normal behavior after submitting a question
* Go to `/chat/?engine=guru-snap`, expect "Chat engine started: Guru SNAP Chat Engine"
   - Expect "TEMP: Replace with generated response once chunks are correct"
* Go to `/chat/?engine=policy-mi`, expect "Chat engine started: Michigan Bridges Policy Manual Chat Engine"
   - Expect "TEMP: Replace with generated response once chunks are correct"

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -4,6 +4,7 @@ from urllib.parse import parse_qs, urlparse
 
 import chainlit as cl
 from src import chat_engine
+from src.format import format_guru_cards
 from src.login import require_login
 
 logger = logging.getLogger(__name__)
@@ -49,8 +50,8 @@ async def on_message(message: cl.Message) -> None:
     engine: chat_engine.ChatEngineInterface = cl.user_session.get("chat_engine")
     try:
         result = engine.on_message(question=message.content)
-        answer = engine.format_answer_message(result)
-        await cl.Message(content=answer).send()
+        msg_content = result.response + format_guru_cards(result.chunks)
+        await cl.Message(content=msg_content).send()
     except Exception as err:  # pylint: disable=broad-exception-caught
         await cl.Message(
             author="backend",
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -14,7 +14,7 @@ require_login()
 @cl.on_chat_start
 async def start() -> None:
     engine_id = engine_url_query_value()
-    logger.info("engine_id: %s", engine_id)
+    logger.info("Engine ID: %s", engine_id)
     engine = chat_engine.create_engine(engine_id)
     if not engine:
         await cl.Message(
@@ -35,7 +35,7 @@ async def start() -> None:
 
 def engine_url_query_value() -> str:
     url = cl.user_session.get("http_referer")
-    logger.debug("URL: %s", url)
+    logger.debug("Referer URL: %s", url)
 
     # Using this suggestion: https://github.com/Chainlit/chainlit/issues/144#issuecomment-2227543547
     parsed_url = urlparse(url)
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -25,7 +25,7 @@ async def start() -> None:
         return
 
     cl.user_session.set("chat_engine", engine)
-    await engine.on_start()
+    engine.on_start()
     await cl.Message(
         author="backend",
         metadata={"engine": engine_id},
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -25,7 +25,6 @@ async def start() -> None:
         return
 
     cl.user_session.set("chat_engine", engine)
-    engine.on_start()
     await cl.Message(
         author="backend",
         metadata={"engine": engine_id},
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -47,7 +47,15 @@ async def on_message(message: cl.Message) -> None:
     logger.info("Received: %r", message.content)
 
     engine: chat_engine.ChatEngineInterface = cl.user_session.get("chat_engine")
-    results = engine.on_message(question=message.content, cl_message=message)
-    answer = engine.format_answer_message(results)
-
-    await cl.Message(content=answer).send()
+    try:
+        results = engine.on_message(question=message.content, cl_message=message)
+        answer = engine.format_answer_message(results)
+        await cl.Message(content=answer).send()
+    except Exception as err:  # pylint: disable=broad-exception-caught
+        await cl.Message(
+            author="backend",
+            metadata={"error_class": err.__class__.__name__, "error": str(err)},
+            content=f"{err.__class__.__name__}: {err}",
+        ).send()
+        # Re-raise error to have it in the logs
+        raise err
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -49,8 +49,8 @@ async def on_message(message: cl.Message) -> None:
 
     engine: chat_engine.ChatEngineInterface = cl.user_session.get("chat_engine")
     try:
-        results = engine.on_message(question=message.content, cl_message=message)
-        answer = engine.format_answer_message(results)
+        result = engine.on_message(question=message.content, cl_message=message)
+        answer = engine.format_answer_message(result)
         await cl.Message(content=answer).send()
     except Exception as err:  # pylint: disable=broad-exception-caught
         await cl.Message(
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -2,12 +2,9 @@ import logging
 from urllib.parse import parse_qs, urlparse
 
 import chainlit as cl
-import src.adapters.db as db
+from src import chat_engine
 from src.format import format_guru_cards
-from src.generate import generate
 from src.login import require_login
-from src.retrieve import retrieve
-from src.shared import get_embedding_model
 
 logger = logging.getLogger(__name__)
 
@@ -16,8 +13,24 @@ require_login()
 
 @cl.on_chat_start
 async def start() -> None:
-    chat_engine = engine_url_query_value()
-    print("chat_engine", chat_engine)
+    engine_id = engine_url_query_value()
+    logger.info("engine: %s", engine_id)
+    engine = chat_engine.create_engine(engine_id)
+    if not engine:
+        await cl.Message(
+            author="backend",
+            metadata={"engine": engine_id},
+            content=f"Available engines: {chat_engine.available_engines()}",
+        ).send()
+        return
+
+    cl.user_session.set("chat_engine", engine)
+    engine.on_start()
+    await cl.Message(
+        author="backend",
+        metadata={"engine": engine_id},
+        content=f"Chat engine started: {engine.name}",
+    ).send()
 
 
 def engine_url_query_value() -> str:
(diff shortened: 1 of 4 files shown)


app/src/chat_engine.py
```diff
@@ -0,0 +1,146 @@
...
+    return chat_engine_class()
+
+
+# Subclasses of ChatEngineInterface can be extracted into a separate file if it gets too large
+class GuruBaseEngine(ChatEngineInterface):
+    use_snap_dataset_default = False
+    use_multiprogram_dataset_default = False
+
+    async def on_start(self) -> dict:
+        chat_settings = cl.ChatSettings(
```
KevinJBoyer, line 64: Would recommend deferring this to another ticket that implements these settings (and leaving a TODO comment here that references the ticket)

app/src/chat_engine.py
```diff
@@ -0,0 +1,146 @@
...
+
+    chat_engine_class = next(
+        engine_class
+        for engine_class in all_subclasses(ChatEngineInterface)
+        if hasattr(engine_class, "engine_id") and engine_class.engine_id == engine_id
+    )
+    return chat_engine_class()
+
+
+# Subclasses of ChatEngineInterface can be extracted into a separate file if it gets too large
```
KevinJBoyer, line 58: 👍  yeah I imagine in the future we'd want to put these under something like /app/src/engines

app/src/chainlit.py
```diff
@@ -1,32 +1,61 @@
...
 require_login()
 
 
-@cl.on_message
-async def main(message: cl.Message) -> None:
-    logger.info(f"Received: {message.content!r}")
+@cl.on_chat_start
+async def start() -> None:
+    engine_id = engine_url_query_value()
+    logger.info("engine_id: %s", engine_id)
```
KevinJBoyer, line 16:
```suggestion
    logger.info("Engine ID: %s", engine_id)
```
to make it human readable, or `logger.info(f"{engine_id=}")` for consistent formatting of variable name / value pairs

app/src/chainlit.py
```diff
@@ -1,32 +1,61 @@
...
-            db_session,
-            get_embedding_model(),
-            message.content,
-        )
 
-    response = generate(message.content, context=chunks)
-    formatted_guru_cards = format_guru_cards(chunks)
+def engine_url_query_value() -> str:
+    url = cl.user_session.get("http_referer")
+    logger.debug("URL: %s", url)
```
KevinJBoyer, line 37:
```suggestion
    logger.debug("Referer URL: %s", url)
```
(keeping the logs slightly more precise)

app/src/chat_engine.py
```diff
@@ -0,0 +1,146 @@
...
+    @abstractmethod
+    def on_message(self, question: str, cl_message: cl.Message) -> dict:
+        pass
+
+    @abstractmethod
+    def format_answer_message(self, results: dict) -> str:
+        pass
+
+
+def all_subclasses(cls: Type) -> set:
```
KevinJBoyer, line 34: Bit messy but I think you should be able to do something like:
```
from typing import Type, TypeVar

T = TypeVar('T')

def all_subclasses(cls: Type[T]) -> set[Type[T]]:

File: app/src/chainlit.py [hunk 6]
```diff
@@ -1,32 +1,61 @@
...
+        ).send()
+        return
+
+    cl.user_session.set("chat_engine", engine)
+    await engine.on_start()
+    await cl.Message(
+        author="backend",
+        metadata={"engine": engine_id},
+        content=f"Chat engine started: {engine.name}",
+    ).send()
```
KevinJBoyer, line 32: Do we want to show this to the user?

KevinJBoyer (commented):
LGTM! Lots of optional nits around style, with two bigger pieces of feedback:

 - Suggest we defer adding the user settings to a later ticket, I'm wary of adding a large chunk of functionality without that being fully wired in
  - I think the interface for ChatEngineInterface should be `on_message(message: str) -> str` rather than `on_message` and `format_answer_message`

app/src/chat_engine.py
KevinJBoyer: (and a sidenote that this is a really slick way of automatically keeping the list of available engines up to date 👍 )

app/src/chainlit.py
yoomlam: I prefer some feedback rather than silently failing and the user is waiting for an answer that never arrives.

app/src/chat_engine.py
```diff
@@ -0,0 +1,146 @@
...
+    return {cls}.union(s for c in cls.__subclasses__() for s in all_subclasses(c))
+
+
+def available_engines() -> list[str]:
+    return [
+        engine_class.engine_id
+        for engine_class in all_subclasses(ChatEngineInterface)
+        if hasattr(engine_class, "engine_id") and engine_class.engine_id
```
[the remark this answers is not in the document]
yoomlam: Without it, I get:
```
AttributeError: type object 'ChatEngineInterface' has no attribute 'engine_id'
```

app/src/chat_engine.py
yoomlam: Why is that? The ticket that this PR is intended to resolve (https://navalabs.atlassian.net/browse/DST-327) requires that the appropriate configuration be set for the specified engine.

yoomlam: I removed `temperature` and `retrieve_k` settings. The remaining settings are used in a test.

app/src/chainlit.py
KevinJBoyer: That makes sense for the earlier message about failure, but this is a notice that the chat engine was started.

yoomlam:
> * I think the interface for ChatEngineInterface should be `on_message(message: str) -> str` rather than `on_message` and `format_answer_message`

My reason for splitting them is to separate data (model) from presentation (view). `on_message` returns an `OnMessageResult` in a UI-independent representation, while `format_answer_message` returns a string formatted for the user. Additionally, there may be other useful structured data in `OnMessageResult` that is not in the `str` returned from `on_message(message: str) -> str`.

app/src/chainlit.py
yoomlam: Oops, I was thinking of the wrong code! 
We can replace this with something else or remove it.

app/src/chat_engine.py
yoomlam: https://github.com/navapbc/labs-decision-support-tool/actions/runs/9958216005/job/27512064988?pr=25#step:4:16

app/src/chat_engine.py
```diff
@@ -0,0 +1,146 @@
...
+    async def on_start(self) -> dict:
+        pass
+
+    @abstractmethod
+    def on_message(self, question: str, cl_message: cl.Message) -> dict:
+        pass
+
+    @abstractmethod
+    def format_answer_message(self, results: dict) -> str:
+        pass
```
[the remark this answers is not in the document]
yoomlam: Originally it was just `question` to keep it chainlit-independent. But then I added `chainlit.input_widget` and thought maybe we'd want to include the entire `cl.Message` for reference.
I'll remove `cl.Message`.

app/src/chat_engine.py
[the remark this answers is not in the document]
KevinJBoyer: Gotcha!

app/src/chat_engine.py
KevinJBoyer: I don't think the two booleans here should be part of user settings at all. These are options presented to the user in the UI. We don't want them to be able to override the chat engine they've selected via the URL.

app/src/chat_engine.py
KevinJBoyer: Ah, unfortunately it looks like my suggestion doesn't work for abstract classes. Yet another reason to wish for a better typing system for Python 😭

KevinJBoyer (approved):
> My reason for splitting them is to separate data (model) from presentation (view). on_message returns an OnMessageResult in a UI-independent representation, while format_answer_message returns a string formatted for the user. Additionally, there may be other useful structured data in OnMessageResult that is not in the str returned from on_message(message: str) -> str.

I like separating out model from view, but I disagree with putting that distinction in `chainlit.py::on_message`. You can already see why in this PR:  it's forcing the addition of a new type, `OnMessageResult`, that every chat engine now has to conform with -- but OnMessageResult is really a collection of internal implementation details that can/will/should vary between chat engines. The presence of a generic `OnMessageResult.data` field is a strong signal that chainlit.py::on_message should not be specifying an intermediary type here, because it neither knows enough about the return value from ChatEngineInterface.on_message to specify it, nor is it actually dependent on it.

Approving the PR -- if you feel strongly about keeping the code structured this way, then I would ask that we at minimum remove the `data` field from OnMessageResult before merging and instead make ourselves add optional fields as they're needed to OnMessageResult so that we can explicitly see the how this result varies between engines.

app/src/chat_engine.py
```diff
@@ -0,0 +1,113 @@
...
+    data: Optional[dict] = None
+
+
+class ChatEngineInterface(ABC):
+    engine_id: str
+    name: str
+
+    @abstractmethod
+    async def on_start(self) -> dict:
+        pass
```
KevinJBoyer, line 30: Should this really return None?

```suggestion
    @abstractmethod
    async def on_start(self) -> None:
        pass
```

yoomlam: Yea, I thought about that too. Removing...

yoomlam:
> > My reason for splitting them is to separate data (model) from presentation (view). on_message returns an OnMessageResult in a UI-independent representation, while format_answer_message returns a string formatted for the user. Additionally, there may be other useful structured data in OnMessageResult that is not in the str returned from on_message(message: str) -> str.
> 
> I like separating out model from view, but I disagree with putting that distinction in `chainlit.py::on_message`. You can already see why in this PR: it's forcing the addition of a new type, `OnMessageResult`, that every chat engine now has to conform with -- but OnMessageResult is really a collection of internal implementation details that can/will/should vary between chat engines. The presence of a generic `OnMessageResult.data` field is a strong signal that chainlit.py::on_message should not be specifying an intermediary type here, because it neither knows enough about the return value from ChatEngineInterface.on_message to specify it, nor is it actually dependent on it.
> 
> Approving the PR -- if you feel strongly about keeping the code structured this way, then I would ask that we at minimum remove the `data` field from OnMessageResult before merging and instead make ourselves add optional fields as they're needed to OnMessageResult so that we can explicitly see the how this result varies between engines.

Point taken. My first instinct was to extract `format_answer_message()` into an entirely entity separate from the chat engine. I removed `format_answer_message()`.

ccheng26 (approved):
lgtm

app/src/chainlit.py
```diff
@@ -1,32 +1,62 @@
...
+
+
+@cl.on_message
+async def on_message(message: cl.Message) -> None:
+    logger.info("Received: %r", message.content)
+
+    engine: chat_engine.ChatEngineInterface = cl.user_session.get("chat_engine")
+    try:
+        result = engine.on_message(question=message.content)
+        msg_content = result.response + format_guru_cards(result.chunks)
```
KevinJBoyer, line 53: Clarifying my earlier post: I like separating view + model, which in your original implementation was happening by defining two different functions in GuruBaseEngine -- I would keep those as is, but make GuruBaseEngine.on_message or some other function responsible for calling both. So the control flow looks like:

```python
# chainlit.py
def on_message():
    response = engine.on_message(question=message.content)
    ...
```

```python
# chat_engine.py
class GuruBaseEngine:
    def on_message(message: str) -> str:
        result = process(message)
        return format(result)
        
    def process(message: str) -> type TBD by GuruBaseEngine
        ... # Controller
    
    def format(type TBD) -> str:
        ... # View
```

yoomlam: Yes, that's how I understood your original message. I actually don't want any UI/text formatting in the chat engines -- I see them as backend implementations. Any formatting should be done by a separate entity. So it's good that you convinced me to remove chainlit imports from `chat_engine.py`, which convinced me to restore `format_guru_cards()` back in `chainlit.py`. 

I think a chat engine can specify the output formatting as a configuration value, and some other (view-centric) entity can perform that formatting.

PR #26, opened 2024-07-16
fix: set uvicorn worker to 1

ccheng26:
## Changes

> change amount of worker #'s to 1

## Context for reviewers

> Chainlit uses Socket.io for backend server connections; multiple workeres are currently unsupported by gunicorn
https://flask-socketio.readthedocs.io/en/latest/deployment.html#gunicorn-web-server

## Testing

> Run `make start`
> Navigate to `localhost:8000/chat` and enter a query into the chatbot, response should be returned


yoomlam (changes requested):
Also revert docker-compose to use `uvicorn` so that hot-loading is restored

yoomlam (approved):

PR #27, opened 2024-07-16
feat: Enable ingest of distinct datasets

yoomlam:
## Ticket

Partly satisfies https://navalabs.atlassian.net/browse/DST-328 by setting up the DB columns and ingestion process. Completing the ticket will be done in a separate PR, which will flush out the `GuruSnapEngine` implementation and query the DB using the new columns.

## Changes

Enable ingest of distinct datasets, i.e. Guru SNAP and Guru Multiprogram.

## Context for reviewers

See ticket

## Testing

```
# Reset the database and run migrations
make db-recreate

# Ingest SNAP cards
make ingest-guru-cards DATASET_ID=guru-snap BENEFIT_PROGRAM=SNAP BENEFIT_REGION=Michigan FILEPATH=/app/guru_cards_setA.json
```

Check the DB values (175 rows):
image

```
# Ingest Multiprogram cards
make ingest-guru-cards DATASET_ID=guru-multiprogram BENEFIT_PROGRAM=multiprogram BENEFIT_REGION=Michigan FILEPATH=/app/guru_cards_setB.json
```
Check the DB values (175 + 450 rows), noting the different `dataset` and `program` values:
image


app/.gitignore
```diff
@@ -31,7 +31,7 @@ coverage.*
 poetry-installer-error-*.log
 
 # Cached models
-models/
+/models/
```
yoomlam, line 34: Prefix with `/` to avoid ignoring `src/db/models`

app/src/ingest_guru_cards.py
```diff
@@ -51,18 +54,21 @@ def _ingest_cards(
...
+    dataset_id = sys.argv[1]
+    guru_cards_filepath = sys.argv[2]
 
-    logger.info(f"Processing Guru cards at {guru_cards_filepath}")
+    logger.info(f"Processing Guru cards {dataset_id} at {guru_cards_filepath}")
 
     embedding_model = get_embedding_model()
 
+    # TODO: Add `region` as a argv element or have a better way to set document attributes
+    doc_attribs: dict[str, str | None] = {"program": dataset_id, "region": "Michigan"}
```
yoomlam, line 69: Hardcoding the value for `region` for now.

yoomlam, line 69: @KevinJBoyer  Should `dataset_id` be renamed to `program_id`:
```suggestion
    # TODO: Add `region` as a argv element or have a better way to set document attributes
    doc_attribs: dict[str, str | None] = {"program": program_id, "region": "Michigan"}
```
or is there a mapping from `dataset_id` to `program` and `region`?

app/tests/src/db/models/factories.py
```diff
@@ -80,6 +83,4 @@ class Meta:
     tokens = factory.LazyAttribute(
         lambda o: len(MockSentenceTransformer().tokenizer.tokenize(o.content))
     )
-    mpnet_embedding = factory.LazyAttribute(
-        lambda o: MockSentenceTransformer().encode(o.content)
-    )
+    mpnet_embedding = factory.LazyAttribute(lambda o: MockSentenceTransformer().encode(o.content))
```
yoomlam, line 86: `make format` made this change, but it shouldn't be needed since `make lint` didn't fail. How can we update `make format` to prevent making this change?

app/Makefile
```diff
@@ -228,10 +228,15 @@ help: ## Prints the help documentation and info about each command
 # ETL
 ##################################################
 
+check-dataset-id:
+ifndef DATASET_ID
+	$(error DATASET_ID is undefined)
+endif
+
```
KevinJBoyer, line 235: Suggesting changing this to `program` for consistency

Also threading arbitrarily, we should also update the example in the root README.md to include this parameter.

app/src/db/models/document.py
```diff
@@ -21,6 +21,12 @@ class Document(Base, IdMixin, TimestampMixin):
         "Chunk", back_populates="document", cascade="all, delete"
     )
 
+    # Domain-specific columns follow
+    # benefit program
+    program: Mapped[str | None]
+    # geographical region of the benefit program
+    region: Mapped[str | None]
```
KevinJBoyer, line 28: Not something I feel super strongly about, but what do you think about making these non-optional?

app/src/ingest_guru_cards.py
KevinJBoyer: Looks like this may be out of date with your latest change, but I agree with using `program`.

app/tests/src/db/models/factories.py
```diff
@@ -69,6 +70,8 @@ class Meta:
 
     name = factory.Faker("word")
     content = factory.Faker("text")
+    program = random.choice([None, "SNAP", "Medicaid", "TANF"])
+    region = random.choice([None, "MI", "MD", "PA"])
```
KevinJBoyer, line 74: You'll want to use `factory.Faker('random_choices', elements=['XYZ', 'ABC', '123', '456'])` here -- random.choice will be evaluated once as Python reads the class definition, so every subsequent call to the factory will have the same program/region.

nit: may want to make this consistent with the example usage ("guru-snap", "Michigan"). If you wanted, you could enforce this at the DB level by making these enum types (example [here](https://github.com/navapbc/template-application-flask/blob/ecc0dd050348632f0ac641581d075edb703303f2/app/src/db/models/user_models.py#L51))

app/tests/src/db/models/factories.py
KevinJBoyer: Weird, also this wasn't a change in this PR either. I wonder what happened that caused black to put this all in one line? (I'm seeing the same thing on my machine.) If I put it either way on my machine, `make format` does nothing...

KevinJBoyer (approved):
LGTM, with a couple of suggestions around consistency and one update to the factory. Could you also add or update a test in `test_ingest_guru_cars.py`?  (bonus, I'm noticing that `test__ingest_cards_local` is poorly named since it's parameterized to also test S3, so should probably be just `test__ingest_cards`)

app/tests/src/db/models/factories.py
yoomlam: Good catch! Thanks for the solution!

ccheng26 (approved):

app/src/ingest_guru_cards.py
KevinJBoyer: That makes a ton of sense to me, I was seeing the same need but called it `source` rather than dataset in the feedback for policy IDs (I don't have any opinion about the naming here both make sense).

My preference would be try to keep as much of this outside of the code as possible (i.e., not have the mapping of dataset -> program & region hardcoded, but instead make users provide each) to try to separate the data contents (various programs, regions, and dataset names) from the code itself as much as possible-- that keeps the code as flexible and general/generic. E.g., it would save to make any code changes or redeploy the application to ingest a new data set.

yoomlam: I was trying to avoid having a long ingest command but it's acceptable for now. It will look something like:
```sh
make ingest-guru-cards DATASET_ID=policy-mi PROGRAM=SNAP REGION=Michigan FILEPATH=/app/bem-1.pdf
```

app/tests/src/db/models/factories.py
yoomlam: > If you wanted, you could enforce this at the DB level by making these enum types (example [here](https://github.com/navapbc/template-application-flask/blob/ecc0dd050348632f0ac641581d075edb703303f2/app/src/db/models/user_models.py#L51))

Don't think we want to use enum since we're allowing `program` and `region` values to be specified on the command line and be flexible to new values without redeployment, so keeping it as `Text`.

README.md
```diff
@@ -27,9 +27,13 @@ The application supports loading Guru cards from .json files.
 To load a JSON file containing Guru cards in your local environment, from within `/app`:
 
 ```bash
-make ingest-guru-cards DATASET_ID=some_identifier FILEPATH=path/to/some_cards.json
+make ingest-guru-cards DATASET_ID=dataset_identifier BENEFIT_PROGRAM=SNAP BENEFIT_REGION=Michigan FILEPATH=path/to/some_cards.json
 ```
 
+- The same `DATASET_ID` identifier can be used for multiple documents (`FILEPATH`) to represent that the documents belong in the same dataset.
+- Example `BENEFIT_PROGRAM` values include `housing`, `utilities`, `food`, `medical`, `employment`, `SNAP`, `Medicaid`.
```
ccheng26, line 34:
```suggestion
- Example `BENEFIT_PROGRAM` values include `housing`, `utilities`, `food`, `medical`, `employment`, `SNAP`, `Medicaid`, `multi_benefit`, `etc`.
```

ccheng26: Adding this so it doesn't appear to be a restrictive list.

yoomlam: I figured "Example" and "include" would convey that it's not a complete list.

app/src/ingest_guru_cards.py
```diff
@@ -51,18 +54,33 @@ def _ingest_cards(
 
 
 def main() -> None:
-    if len(sys.argv) < 2:
-        logger.warning("No argument for filepath.")
+    if len(sys.argv) < 3:
```
KevinJBoyer, line 57:
```suggestion
    if len(sys.argv) < 5:
```
(I think?)

app/src/ingest_guru_cards.py
```diff
@@ -51,18 +54,33 @@ def _ingest_cards(
...
+    guru_cards_filepath = sys.argv[4]
 
-    logger.info(f"Processing Guru cards at {guru_cards_filepath}")
+    logger.info(
+        f"Processing Guru cards {dataset_id} at {guru_cards_filepath} for {benefit_program} in {benefit_region}"
+    )
 
     embedding_model = get_embedding_model()
 
+    # TODO: Add `region` as a argv element or have a better way to set document attributes
```
KevinJBoyer, line 76: Is this to do now addressed?

app/tests/src/test_ingest_guru_cards.py
```diff
@@ -39,15 +39,22 @@ def guru_s3_file(mock_s3_bucket_resource, sample_cards):
...
 
     if file_location == "local":
-        _ingest_cards(db_session, mock_embedding, guru_local_file)
+        _ingest_cards(db_session, mock_embedding, guru_local_file, doc_attribs)
     else:
-        _ingest_cards(db_session, mock_embedding, guru_s3_file)
+        _ingest_cards(db_session, mock_embedding, guru_s3_file, doc_attribs)
 
     documents = db_session.execute(select(Document).order_by(Document.name)).scalars().all()
     assert len(documents) == 3
```
KevinJBoyer, line 60: Suggest add these assertions to verify doc_attribs is being processed as we expect:
```suggestion
    assert len(documents) == 3
    assert set(d.dataset for d in documents) == {"test_dataset"}
    assert set(d.program for d in documents) == {"test_benefit_program"}
    assert set(d.region for d in documents) == {"Michigan"}
```

KevinJBoyer (approved):
LGTM!

app/src/ingest_guru_cards.py
yoomlam: Yup! Deleted line.

PR #28, opened 2024-07-17
feat: Set default chat engine to 'guru-snap'

yoomlam:
## Ticket

As requested by Ryan, set the default chat engine to 'guru-snap'.

## Changes
Change default chat engine to 'guru-snap'.

## Testing

Navigate to `/chat` (without the `engine` query parameter), expect "Chat engine started: Guru SNAP Chat Engine".

Navigate to `/chat/?engine=someNonExistingEngineId` to see available engines.

### DIFF
--- a/app/gunicorn.conf.py
+++ b/app/gunicorn.conf.py
@@ -7,11 +7,10 @@ Attributes:
 
 For more information, see https://docs.gunicorn.org/en/stable/configure.html
 """
-import os
 
-from src.app_config import AppConfig
+from src import shared
 
-app_config = AppConfig()
+app_config = shared.get_app_config()
 
 # Since the `-b 0.0.0.0:8000` argument is used when running in the Docker environment,
 # this bind variable is only used when not using Docker
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from urllib.parse import parse_qs, urlparse
 
 import chainlit as cl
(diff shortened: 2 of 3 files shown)


app/src/chainlit.py
```diff
@@ -40,7 +40,7 @@ def engine_url_query_value() -> str:
     # Using this suggestion: https://github.com/Chainlit/chainlit/issues/144#issuecomment-2227543547
     parsed_url = urlparse(url)
     qs = parse_qs(parsed_url.query)
-    return qs.get("engine", [os.environ.get("CHAT_ENGINE", "default_engine")])[0]
+    return qs.get("engine", [os.environ.get("CHAT_ENGINE", "guru-snap")])[0]
```
KevinJBoyer, line 43: Sorry I missed this in an earlier review -- we should set this in AppConfig and put a default value there, then use AppConfig().chat_engine to retrieve the configuration value (consistent with how the rest of the app is configured)_. AppConfig uses Pydantic to auto-pull environment variables so it can still be overridden by setting CHAT_ENGINE in the environment if needed.

app/gunicorn.conf.py
```diff
@@ -7,11 +7,10 @@
 
 For more information, see https://docs.gunicorn.org/en/stable/configure.html
 """
-import os
 
-from src.app_config import AppConfig
+from src import shared
 
-app_config = AppConfig()
+app_config = shared.get_app_config()
```
yoomlam, line 13: Modified to use shared.get_app_config() consistently

ccheng26 (approved):

app/gunicorn.conf.py
KevinJBoyer: Good idea -- looks like [there is a way](https://stackoverflow.com/questions/8212053/private-constructor-in-python) but it's not simple

KevinJBoyer (approved):

PR #29, opened 2024-07-17
feat: Retrieve from database based on filters

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-328
follow-on to PR #27

## Changes

Flush out the `GuruSnapEngine` implementation by enabling query the DB using filters (SQL `WHERE`) against the new `dataset`, `program`, and `region` columns.

## Testing

Grab datasets from [our S3 bucket](https://us-east-1.console.aws.amazon.com/s3/buckets/decision-support-tool-app-dev?bucketType=general&region=us-east-1&tab=objects#), and move them to the `app` folder. ([dataset descriptions](https://nava.slack.com/archives/C06ETE82UHM/p1717787155567219?thread_ts=1717783522.574479&cid=C06ETE82UHM))

```
# Reset the database and run migrations
make db-recreate

# Ingest guru-snap dataset, which will be used by the guru-snap engine
make ingest-guru-cards DATASET_ID=guru-snap BENEFIT_PROGRAM=SNAP BENEFIT_REGION=Michigan FILEPATH=/app/guru_cards_setA.json

# Ingest guru-multiprogram dataset, which will be used by the guru-multiprogram engine
make ingest-guru-cards DATASET_ID=guru-multiprogram BENEFIT_PROGRAM=multiprogram BENEFIT_REGION=Michigan FILEPATH=/app/guru_cards_setB.json
```

1. Go to `/chat/` or `/chat/?engine=guru-snap`; expect "Chat engine started: Guru SNAP Chat Engine"
1. Ask a SNAP-related question like "SNAP eligibility" and expect Guru cards related to SNAP
2. Ask a non-SNAP-related question like "energy assistance" and expect **irrelevant** Guru cards (b/c it's only querying against the SNAP Guru cards)

Unchanged:
1. Go to `/chat/?engine=guru-multiprogram`; expect "Chat engine started: Guru Multi-program Chat Engine"
2. Ask a non-SNAP-related question like "energy assistance" and expect **relevant** Guru cards

PR #30, opened 2024-07-17
feat: add similarity score to accordion value

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-333

## Testing

1.  Navigate to `localhost:8000/chat` and enter query
2.  Expand accordion, similarity score value should appear in body of text

Screenshot 2024-07-17 at 5 17 27 PM

1. Navigate to LiteralAI page 
2. click on Threads
3. click into thread query and scroll to metadata, title with similarity score should appear in metadata
Screenshot 2024-07-17 at 5 19 39 PM


app/src/chat_engine.py
```diff
@@ -52,14 +53,16 @@ def create_engine(engine_id: str) -> ChatEngineInterface | None:
 class GuruBaseEngine(ChatEngineInterface):
     def on_message(self, question: str) -> OnMessageResult:
         with db.PostgresDBClient().get_session() as db_session:
-            chunks = retrieve(
+            chunks = retrieve_with_scores(
```
yoomlam, line 58: Crux of my feedback: `retrieve_with_scores()` returns `Sequence[Row[Tuple[Chunk, float]]]`, whereas `retrieve()` returns a basic `Sequence[Chunk]`. This means `chunks` is now a misnomer -- it should be `chunks_with_scores`. To facilitate future development, anywhere you use `chunk[1]` should be renamed `score`, and `chunk[0]` is the original `chunk`.

app/src/chat_engine.py
```diff
@@ -17,6 +17,7 @@
 class OnMessageResult:
     response: str
     chunks: Sequence[Chunk]
```
yoomlam: To simplify, we can always use `retrieve_with_scores()` instead of `retrieve()` and store this:
```suggestion
    chunks_with_scores: dict[Chunk, float]
```
cc: @KevinJBoyer

app/src/chainlit.py
```diff
@@ -51,7 +51,10 @@ async def on_message(message: cl.Message) -> None:
     try:
         result = engine.on_message(question=message.content)
         msg_content = result.response + format_guru_cards(result.chunks)
-        await cl.Message(content=msg_content).send()
+        await cl.Message(
+            content=msg_content,
+            metadata={chunk[0].document.name: chunk[1] for chunk in result.chunks},
```
yoomlam: When reading these code changes from the top down, I didn't realize `chunk[1]` was the score.

yoomlam (changes requested):
The approach looks good, but the code needs clean-up to avoid future confusion.

app/src/chat_engine.py
yoomlam: In [my PR, I end up calling](https://github.com/navapbc/labs-decision-support-tool/pull/29/files#diff-576e93798fc73bf515ae76fa5bed4c8df408fc09b7e159cd95646cf12dd7deaaR20) `retrieve_with_scores()` within `retrieve()` anyway.

KevinJBoyer: I wonder if it doesn't make sense to remove `retrieve` entirely and just have `retrieve_with_scores` 🤔

ccheng26: I'm thinking of going with the approach for removing `retrieve` entirely for more predictably

app/src/chat_engine.py
```diff
@@ -3,10 +3,12 @@
 from dataclasses import dataclass
 from typing import Sequence
 
+from sqlalchemy import Row
```
yoomlam, line 6: So as not to expose `sqlalchemy.Row` in many files, I would replace `Sequence[Row[tuple[Chunk, float]]]` with `Sequence[ChunkData]` returned by `retrieve_with_scores()`, where `CheckData` is a new `@dataclass` with fields `chunk` and `score`, which provides the option of later adding `document` or `program` if needed. This suggestion (if desirable) can be done a separate PR.

yoomlam (approved):

app/src/chat_engine.py
yoomlam: @KevinJBoyer Thoughts on the above comment? Shall I create a PR to implement it?

KevinJBoyer: Makes sense to me, though I'd have a slight preference for calling it ChunksWithScores or something like that vs. data. Alternatively, we could define the type directly like `ChunksWithScores = Row[tuple[Chunk, float]]` -- there might be disadvantages to that approach vs a dataclass though

yoomlam: https://github.com/navapbc/labs-decision-support-tool/pull/32

PR #31, opened 2024-07-22
feat: Add ingest-policy-pdfs command to print out PDF file list

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-341

## Changes
- added cmd to iterate through directory and print pdf titles

## Testing
Run 
```
make ingest-policy-pdfs DATASET_ID=dataset_identifier BENEFIT_PROGRAM=SNAP BENEFIT_REGION=Michigan FILEPATH=path/to/policy.pdf
```
in directory with pdfs, should see a list of pdf titles printed.

Screenshot 2024-07-22 at 3 14 56 PM


app/src/ingest_policy_pdfs.py
yoomlam, line 1: Rather than copy from `ingest_guru_cards.py`, refactor code in `ingest_guru_cards.py` into reusable functions and call them from `ingest_guru_cards.py` and `ingest_policy_pdfs.py`. This will make it easier to tell the differences among ingestion pipelines. This will also reduce the number of test cases, i.e., you don't need to test for local and s3 file locations more than once.

app/src/util/file_util.py
```diff
@@ -34,6 +34,19 @@ def get_file_name(path: str) -> str:
     return os.path.basename(path)
 
 
+def get_files(path: str) -> list[str]:
+    # Return a list of paths to all files in a directory,
+    # whether on local disk or on S3
```
yoomlam, line 39: Convert into a docstring

yoomlam (changes requested):

yoomlam (approved):

PR #33, opened 2024-07-24
refactor: Move db_session and embedding_model args into AppConfig

yoomlam:
## Ticket

From [Slack](https://nava.slack.com/archives/C06DP498D1D/p1721764756834649):

> - [retrieve_with_scores()](https://github.com/navapbc/labs-decision-support-tool/blob/cb9d663d1f575cc06119f561893eff3ee099288e/app/src/retrieve.py#L15) and [_ingest_cards()](https://github.com/navapbc/labs-decision-support-tool/blob/cb9d663d1f575cc06119f561893eff3ee099288e/app/src/ingest_guru_cards.py#L26) take the 2 arguments `db_session` and `embedding_model`.
> - I see these 2 arguments as "[context](https://www.baeldung.com/cs/context-design-pattern)" or "fixtures", and not really arguments that would change with different calls to these functions.
> - Additionally, any retrieval function should use the same `embedding_model` as the ingest function.
> What do y'all think about wrapping these 2 arguments into a context class or moving them into the `AppConfig` class?

## Changes

Move `db_session` and `embedding_model` args into `AppConfig` to be used as a context object.

## Testing

No changes in software behavior.

### DIFF
--- a/app/src/app_config.py
+++ b/app/src/app_config.py
@@ -1,3 +1,7 @@
+from functools import cached_property
+
+from sentence_transformers import SentenceTransformer
+
 import src.adapters.db as db
 from src.util.env_config import PydanticBaseEnvConfig
 
@@ -22,4 +26,10 @@ class AppConfig(PydanticBaseEnvConfig):
     chat_engine: str = "guru-snap"
 
     def db_session(self) -> db.Session:
+        print("\n=====\n========= REAL AppConfig.db_session")
         return db.PostgresDBClient().get_session()
+
+    @cached_property
+    def sentence_transformer(self) -> SentenceTransformer:
+        print("\n=====\n========= REAL AppConfig.sentence_transformer")
+        return SentenceTransformer(self.embedding_model)
--- a/app/src/app_config.py
+++ b/app/src/app_config.py
@@ -1,3 +1,4 @@
+import src.adapters.db as db
 from src.util.env_config import PydanticBaseEnvConfig
 
 
@@ -19,3 +20,6 @@ class AppConfig(PydanticBaseEnvConfig):
     host: str = "127.0.0.1"
     port: int = 8080
     chat_engine: str = "guru-snap"
+
+    def db_session(self) -> db.Session:
+        return db.PostgresDBClient().get_session()
--- a/app/src/app_config.py
+++ b/app/src/app_config.py
@@ -25,11 +25,15 @@ class AppConfig(PydanticBaseEnvConfig):
     port: int = 8080
     chat_engine: str = "guru-snap"
 
+    print("\n=====\n========= Creating REAL AppConfig")
+
     def db_session(self) -> db.Session:
         print("\n=====\n========= REAL AppConfig.db_session")
+        input("UNEXPECTED Press Enter app_config...")
         return db.PostgresDBClient().get_session()
 
     @cached_property
     def sentence_transformer(self) -> SentenceTransformer:
         print("\n=====\n========= REAL AppConfig.sentence_transformer")
+        input("UNEXPECTED Press Enter app_config...")
         return SentenceTransformer(self.embedding_model)
--- a/app/src/app_config.py
+++ b/app/src/app_config.py
@@ -25,15 +25,9 @@ class AppConfig(PydanticBaseEnvConfig):
     port: int = 8080
     chat_engine: str = "guru-snap"
 
-    print("\n=====\n========= Creating REAL AppConfig")
-
     def db_session(self) -> db.Session:
-        print("\n=====\n========= REAL AppConfig.db_session")
-        input("UNEXPECTED Press Enter app_config...")
         return db.PostgresDBClient().get_session()
 
     @cached_property
     def sentence_transformer(self) -> SentenceTransformer:
-        print("\n=====\n========= REAL AppConfig.sentence_transformer")
-        input("UNEXPECTED Press Enter app_config...")
         return SentenceTransformer(self.embedding_model)
--- a/app/src/app_config.py
+++ b/app/src/app_config.py
@@ -1,4 +1,4 @@
-from functools import cache, cached_property
+from functools import cached_property
 
 from sentence_transformers import SentenceTransformer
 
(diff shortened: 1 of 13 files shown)


app/tests/src/test_ingest_guru_cards.py
```diff
@@ -47,14 +46,13 @@ def guru_s3_file(mock_s3_bucket_resource, sample_cards):
 
 
 @pytest.mark.parametrize("file_location", ["local", "s3"])
-def test__ingest_cards(db_session, guru_local_file, guru_s3_file, file_location):
+def test__ingest_cards(app_config, db_session, guru_local_file, guru_s3_file, file_location):
```
yoomlam, line 49: Include the `app_config` fixture to mock the `sentence_transformer` embedding model used by `_ingest_cards()`.

app/tests/src/test_retrieve.py
```diff
@@ -3,11 +3,8 @@
 
 from src.db.models.document import Document
 from src.retrieve import retrieve_with_scores
-from tests.mock.mock_sentence_transformer import MockSentenceTransformer
 from tests.src.db.models.factories import ChunkFactory, DocumentFactory
 
-mock_embedding_model = MockSentenceTransformer()
```
yoomlam, line 9: `mock_embedding_model`/`MockSentenceTransformer` is now provided by the `app_config` fixture via `MockAppConfig.sentence_transformer()`.

app/src/app_config.py
```diff
@@ -19,3 +24,10 @@ class AppConfig(PydanticBaseEnvConfig):
     host: str = "127.0.0.1"
     port: int = 8080
     chat_engine: str = "guru-snap"
+
+    def db_session(self) -> db.Session:
+        return db.PostgresDBClient().get_session()
+
+    @cached_property
+    def sentence_transformer(self) -> SentenceTransformer:
+        return SentenceTransformer(self.embedding_model)
```
yoomlam, line 33: `AppConfig` functions as a context object that provides access to the DB and `sentence_transformer` embedding model instance.

app/src/retrieve.py
```diff
@@ -1,24 +1,22 @@
...
-import src.adapters.db as db
 from src.db.models.document import Chunk, ChunkWithScore, Document
+from src.shared import get_app_config
 
 logger = logging.getLogger(__name__)
 
 
 def retrieve_with_scores(
-    db_session: db.Session,
-    embedding_model: SentenceTransformer,
```
yoomlam, line 15: Remove context arguments, which are now replaced with fields from `get_app_config()`

app/src/shared.py
```diff
@@ -1,15 +1,8 @@
...
 
 
 @cache
 def get_app_config() -> AppConfig:
     return AppConfig()
-
-
-@cache
-def get_embedding_model() -> SentenceTransformer:
-    return SentenceTransformer(get_app_config().embedding_model)
```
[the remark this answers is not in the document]
KevinJBoyer: At this point, should we also move get_app_config to app_config.py and delete this file?

app/src/ingest_guru_cards.py
```diff
@@ -71,15 +70,13 @@ def main() -> None:
...
-    embedding_model = get_embedding_model()
-
     doc_attribs = {
         "dataset": dataset_id,
         "program": benefit_program,
         "region": benefit_region,
     }
-    with db.PostgresDBClient().get_session() as db_session:
-        _ingest_cards(db_session, embedding_model, guru_cards_filepath, doc_attribs)
+    with get_app_config().db_session() as db_session:
```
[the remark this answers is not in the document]
yoomlam: I started to do that but noticed the explicit `db_session.commit()`, so I unsure if we wanted to keep the DB transaction explicit.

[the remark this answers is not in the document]
KevinJBoyer: That's a good point -- defer to you on what you think makes the most sense!

app/src/retrieve.py
```diff
@@ -1,24 +1,22 @@
...
     query: str,
     k: int = 5,
     **filters: Sequence[str] | None,
 ) -> Sequence[ChunkWithScore]:
     logger.info("Retrieving context for %r", query)
 
+    embedding_model = get_app_config().sentence_transformer
```
ccheng26, line 19: nit: does get_app_config() need to be called everytime or can we store the results in a variable?

app/src/ingest_guru_cards.py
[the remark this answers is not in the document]
yoomlam: I defer this to a different PR.

app/src/retrieve.py
yoomlam: Good point!

yoomlam: I had to make more changes that anticipated, especially around login, so test the login mechanism carefully and check the deployment after this merges.

app/src/app_config.py
```diff
@@ -19,3 +25,13 @@ class AppConfig(PydanticBaseEnvConfig):
...
+
+
+app_config = AppConfig()
```
yoomlam, line 37: Assuming `AppConfig` only ever needs to be instantiated once per application startup.

app/src/login.py
```diff
@@ -1,18 +1,17 @@
...
     else:
         return None
 
 
 def require_login() -> None:
-    # In addition to setting GLOBAL_PASSWORD, Chainlit also
-    # requires CHAINLIT_AUTH_SECRET to be set (to sign the
-    # authorization tokens.)
+    # Set GLOBAL_PASSWORD, and also set CHAINLIT_AUTH_SECRET for
+    # Chainlit to sign the authorization tokens.
```
yoomlam, line 14: Clarified wording. `GLOBAL_PASSWORD` is not needed by Chainlit.

app/tests/src/test_login.py
```diff
@@ -1,34 +1,24 @@
...
 
 
 def test_require_login_no_password(monkeypatch):
-    if "GLOBAL_PASSWORD" in os.environ:
-        monkeypatch.delenv("GLOBAL_PASSWORD")
-
-    # Rebuild AppConfig with new environment variables
-    src.shared.get_app_config.cache_clear()
-
+    app_config.global_password = None
```
yoomlam, line 7: Set the `global_password` directly and avoid having to `cache_clear()`. This assumes that `PydanticBaseEnvConfig` correctly sets `global_password` from the env variable, which is a valid assumption since the point of this test is `login.py`.

ccheng26 (approved):
code ran without issues, lgtm

PR #34, opened 2024-07-24
fix: retrieve build date and service env

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-304

## Changes
- Added env variable for build_date, fixed service_name variable to call environment key

## Testing
Screenshot 2024-07-26 at 10 09 33 AM


infra/app/service/main.tf
```diff
@@ -153,6 +153,8 @@ module "service" {
   extra_environment_variables = merge({
     FEATURE_FLAGS_PROJECT = module.feature_flags.evidently_project_name
     BUCKET_NAME           = local.storage_config.bucket_name
+    ENVIRONMENT           = var.environment_name
+    BUILD_DATE            = timestamp()
```
yoomlam, line 157: FYI: `BUILD_DATE` will actually reflect the deploy date.

yoomlam (approved):

infra/app/service/main.tf
ccheng26: I was going by the description in the ticket, do you think we should rename the variable or keep it as is?

yoomlam: Keep it based on the ticket's AC:
> The build_date reflects the date the app was last deployed

PR #36, opened 2024-07-24
feat: Normalize similarity scores of retrieved Guru cards

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-352

Currently scores are not normalized, so their range is unknown. It is common to normalize the scores so that they range from -1 to 1.

## Changes

Normalize similarity scores of retrieved Guru cards by using the `multi-qa-mpnet-base-cos-v1` embedding model, which normalizes the vector saved to the DB.

Also added manual test and confirmed that the `max_inner_product` method returns the same score magnitude as using `sentence_transformers.util.dot_score` used in code at https://huggingface.co/sentence-transformers/multi-qa-mpnet-base-cos-v1.

Scores from the DB query are negated, presumably to reverse the default sort order, so the code reverses the negation when recording the score in `ChunkWithScore`; also updated associated test.

## Testing

```
# Reset the database and run migrations
make db-recreate

# Ingest SNAP cards
make ingest-guru-cards DATASET_ID=guru-snap BENEFIT_PROGRAM=SNAP BENEFIT_REGION=Michigan FILEPATH=/app/guru_cards_setA.json

# Ingest Multiprogram cards
make ingest-guru-cards DATASET_ID=guru-multiprogram BENEFIT_PROGRAM=multiprogram BENEFIT_REGION=Michigan FILEPATH=/app/guru_cards_setB.json
```

Run the application and make a query. The scores should be between -1 and 1, where 1 is maximally relevant.

## Reminder

After this is deployed in the `dev` environment, reminder to clear the DB and re-ingest datasets -- see [instructions](https://nava.slack.com/archives/C06DP498D1D/p1721397891286199?thread_ts=1721396512.937399&cid=C06DP498D1D).

### DIFF
--- a/app/src/retrieve.py
+++ b/app/src/retrieve.py
@@ -2,6 +2,7 @@ import logging
 from typing import Sequence
 
 from sqlalchemy import select
+from sentence_transformers import util
 
 from src.app_config import app_config
 from src.db.models.document import Chunk, ChunkWithScore, Document
@@ -38,6 +39,7 @@ def retrieve_with_scores(
         ).all()
 
         for chunk, score in chunks_with_scores:
-            logger.info(f"Retrieved: {chunk.document.name!r} with score {score}")
+            dot_score = util.dot_score(chunk.mpnet_embedding, query_embedding).item()
+            logger.info(f"Retrieved: {chunk.document.name!r} with score {-score}, dot score: {dot_score}")
 
-        return [ChunkWithScore(chunk, score) for chunk, score in chunks_with_scores]
+        return [ChunkWithScore(chunk, -score) for chunk, score in chunks_with_scores]
--- a/app/src/retrieve.py
+++ b/app/src/retrieve.py
@@ -2,7 +2,6 @@ import logging
 from typing import Sequence
 
 from sqlalchemy import select
-from sentence_transformers import util
 
 from src.app_config import app_config
 from src.db.models.document import Chunk, ChunkWithScore, Document
@@ -39,7 +38,9 @@ def retrieve_with_scores(
         ).all()
 
         for chunk, score in chunks_with_scores:
-            dot_score = util.dot_score(chunk.mpnet_embedding, query_embedding).item()
-            logger.info(f"Retrieved: {chunk.document.name!r} with score {-score}, dot score: {dot_score}")
+            # Confirmed that the `max_inner_product` method returns the same score as using sentence_transformers.util.dot_score
+            # used in code at https://huggingface.co/sentence-transformers/multi-qa-mpnet-base-cos-v1
+            logger.info(f"Retrieved: {chunk.document.name!r} with score {-score}")
 
+        # Scores from the DB query are negated, presumably to reverse the default sort order
         return [ChunkWithScore(chunk, -score) for chunk, score in chunks_with_scores]
--- a/app/tests/mock/test_mock_sentence_transformer.py
+++ b/app/tests/mock/test_mock_sentence_transformer.py
@@ -29,14 +29,15 @@ def test_mock_sentence_transformer():
     assert dot_product(medium_text, medium_text) > dot_product(medium_text, short_text)
 
 
-# This test doesn't normally run with the other tests because it requires downloading large embedding models.
-# To run this test, remove the `manual_` prefix so that the function begins with `test_`.
-# Run the test locally: pytest tests/mock/test_mock_sentence_transformer.py --capture=no -k test_sentence_transformer
+# This test is skipped because it requires downloading large embedding models but is good for manual testing.
+# To run this test, comment out the `@pytest.mark.skip ...` line, then run the test locally with:
+# pytest tests/mock/test_mock_sentence_transformer.py --capture=no -k test_sentence_transformer
 # The embedding models will be downloaded automatically to ~/.cache/huggingface/hub, if it does not already exist.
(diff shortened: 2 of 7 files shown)


docs/app/getting-started.md
```diff
@@ -26,7 +26,7 @@ A very simple [docker-compose.yml](/docker-compose.yml) has been included to sup
 **Note:** Run everything from within the `/app` folder:
 
 1. Set up an (empty) local secrets file: `touch .env` and copy the provided example Docker override: `cp ../docker-compose.override.yml.example ../docker-compose.override.yml`
-2. Download the `multi-qa-mpnet-base-dot-v1` model into the `models` directory: `git clone https://huggingface.co/sentence-transformers/multi-qa-mpnet-base-dot-v1 models/multi-qa-mpnet-base-dot-v1`
+2. Download the embedding model into the `models` directory: `git clone https://huggingface.co/sentence-transformers/multi-qa-mpnet-base-cos-v1 models/multi-qa-mpnet-base-cos-v1`
```
yoomlam, line 29: Can we eliminate this step by doing [this](https://stackoverflow.com/a/69717491/23458508)?

app/src/retrieve.py
```diff
@@ -38,6 +38,9 @@ def retrieve_with_scores(
         ).all()
 
         for chunk, score in chunks_with_scores:
-            logger.info(f"Retrieved: {chunk.document.name!r} with score {score}")
+            # Confirmed that the `max_inner_product` method returns the same score as using sentence_transformers.util.dot_score
+            # used in code at https://huggingface.co/sentence-transformers/multi-qa-mpnet-base-cos-v1
+            logger.info(f"Retrieved: {chunk.document.name!r} with score {-score}")
 
-        return [ChunkWithScore(chunk, score) for chunk, score in chunks_with_scores]
+        # Scores from the DB query are negated, presumably to reverse the default sort order
```
KevinJBoyer, line 45: Yup, that's exactly why

app/tests/mock/test_mock_sentence_transformer.py
```diff
@@ -24,3 +27,28 @@ def dot_product(v1, v2):
     assert dot_product(long_text, long_text) > dot_product(long_text, medium_text)
     assert dot_product(long_text, medium_text) > dot_product(long_text, short_text)
     assert dot_product(medium_text, medium_text) > dot_product(medium_text, short_text)
+
```
KevinJBoyer, line 30: Should we update the lines above this to use utils.dot_score instead of the custom dot product function?

app/tests/mock/test_mock_sentence_transformer.py
```diff
@@ -24,3 +27,28 @@ def dot_product(v1, v2):
...
+
+# This test doesn't normally run with the other tests because it requires downloading large embedding models.
+# To run this test, remove the `manual_` prefix so that the function begins with `test_`.
+# Run the test locally: pytest tests/mock/test_mock_sentence_transformer.py --capture=no -k test_sentence_transformer
```
KevinJBoyer, line 34: You can use `@pytest.mark.skip(reason="some reason")` for this functionality

app/tests/mock/test_mock_sentence_transformer.py
```diff
@@ -24,3 +27,28 @@ def dot_product(v1, v2):
...
+# The embedding models will be downloaded automatically to ~/.cache/huggingface/hub, if it does not already exist.
+@pytest.mark.parametrize(
+    "embedding_model", ["multi-qa-mpnet-base-cos-v1", "multi-qa-mpnet-base-dot-v1"]
+)
+def manual_test_sentence_transformer(embedding_model):
```
KevinJBoyer, line 39: Not following the intent here -- do we want to merge this code into the repo?

KevinJBoyer (approved):
LGTM w/ one question around adding that test

docs/app/getting-started.md
KevinJBoyer: Looks like we potentially could!

app/tests/mock/test_mock_sentence_transformer.py
yoomlam: I kinda like that we've tested calculating the score in different ways:
- `dot_product` uses basic math
- `util.dot_score` uses tensor math (linear algebra)
- `.max_inner_product` uses pgvector's sqlalchemy implementation

These different methods could be useful to validate scores later.
With that said, I'm ambivalent since it's easy enough to replicate this work.

app/tests/mock/test_mock_sentence_transformer.py
yoomlam: Yes, I'd like to, in case we want to test other embedding models. Alternatively, I can move it to a regular file that is only used for testing.

ccheng26 (approved):
code ran and working as expected

app/tests/mock/test_mock_sentence_transformer.py
KevinJBoyer: Can you share more about what this is intended to test?

yoomlam: It's not a unit test so I'll move it into a separate file outside of `tests/`.
It exercises different embedding models and calculates scores from the embedding vectors. I used the scores to confirm/compare against those of pgvector's `max_inner_product`.

PR #37, opened 2024-07-26
feat: Add retrieval and docs_shown thresholds

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-356

Currently, the chatbot has 1 threshold (retrieval-k) that controls both:
- retrieval-k: the number of retrieved docs (e.g., Guru cards) that gets sent to the LLM for it to generate the free-text response, and
- the retrieved docs that gets shown to the user (which I’ll call max-num-docs-shown)

We want to be able to control these 2 thresholds independently because:
- we may want to have a high retrieval-k to give as much context as possible to the LLM
- but limit max-num-docs-shown to the user

We also want to control the docs shown based on the retrieval/similarity/relevance score, so also add a docs-shown-min-score, which will also limit the number of docs shown, i.e., a doc will only be shown if it meets the docs-shown-min-score AND does not exceed the max-num-docs-shown.

## Changes

* Adds the following thresholds as settings in the UI:
   - `retrieval_k`: Number of documents to retrieve for generating LLM response
   - `retrieval_k_min_score`: Minimum document score required for generating LLM response
   - `docs_shown_max_num`: Maximum number of retrieved documents to show in the UI
   - `docs_shown_min_score`: Minimum document score required to show document in the UI
* Have these configuration settings affect the behavior of the chatbot

## Context for reviewers

Ryan and team will adjust these settings to determined appropriate thresholds for each chat engine.

## Testing

Start the chatbot, repeatly change a setting in the Chainlit UI, submit a question, then expect the cited Guru cards to satisfy the adjusted settings. Examples include:
- high retrieval_k but lower docs_shown_max_num should show no more than docs_shown_max_num
- low retrieval_k but higher docs_shown_max_num should show no more than retrieval_k

Examine the scores and change the docs_shown_min_score setting value to somewhere in the middle of the score range; rerun the query; and expect fewer documents.

Select a very low or zero retrieval_k or retrieval_k_min_score; expect the generated answer to not use any documents, and no docs would be shown in the UI.

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -6,7 +6,7 @@ from urllib.parse import parse_qs, urlparse
 import chainlit as cl
 from chainlit.input_widget import InputWidget, Slider
 from src import chat_engine
-from src.app_config import UserConfig, app_config
+from src.app_config import app_config
 from src.format import format_guru_cards
 from src.login import require_login
 
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -48,9 +48,9 @@ def _init_chat_engine(engine_id: str) -> ChatEngineInterface | None:
 
 async def _init_chat_settings(engine: ChatEngineInterface) -> dict[str, Any]:
     input_widgets: list[InputWidget] = [
-        factory(engine.user_config)
-        for attrib_name, factory in _WIDGET_FACTORIES.items()
-        if hasattr(engine.user_config, attrib_name)
+        _WIDGET_FACTORIES[setting_name](getattr(engine, setting_name))
+        for setting_name in engine.user_settings
+        if setting_name in _WIDGET_FACTORIES
     ]
     settings = await cl.ChatSettings(input_widgets).send()
     logger.info("Initialized settings: %s", pprint.pformat(settings, indent=4))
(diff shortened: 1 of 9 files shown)


app/tests/src/test_format.py
```diff
@@ -1,22 +1,34 @@
...
-def test_format_guru_cards_with_score(app_config, db_session, enable_factory_create):
+def test_format_guru_cards_with_score(monkeypatch, app_config, db_session, enable_factory_create):
     db_session.execute(delete(Document))
 
-    chunks_with_scores = _get_chunks_with_scores()
-    html = format_guru_cards(chunks_with_scores)
-    assert "accordion-1" in html
+    monkeypatch.setenv("DOCS_SHOWN_MIN_SCORE", "0.0")
+    app_config.reinit()
+    user_config = app_config.create_user_config(retrieval_k=2)
```
yoomlam, line 27: Since we now have the `UserConfig` class, if we don't want to add the `DynamicAppConfig` class, we can use `app_config.create_user_config(retrieval_k=2)` or `user_config.retrieval_k=2` to override *user* configuration settings for unit tests. Note that *non-user* configuration settings in AppConfig would not be reconfigurable without `DynamicAppConfig`.

app/src/chainlit.py
```diff
@@ -25,14 +30,78 @@ async def start() -> None:
...
         return
 
-    cl.user_session.set("chat_engine", engine)
+    settings = await _init_chat_settings(engine)
     await cl.Message(
         author="backend",
-        metadata={"engine": engine_id},
-        content=f"Chat engine started: {engine.name}",
+        metadata={"engine": engine_id, "settings": str(settings)},
+        content=f"{engine.name} started with settings:\n{pprint.pformat(settings, indent=3)}",
```
yoomlam, line 37: Printing the settings for internal team use. We can remove this when piloting with actual users.

app/src/chat_engine.py
```diff
@@ -21,6 +21,17 @@ class ChatEngineInterface(ABC):
     engine_id: str
     name: str
 
+    # Thresholds that determine which retrieved documents are shown in the UI
+    docs_shown_max_num: int = 5
+    docs_shown_min_score: float = 0.65
```
yoomlam, line 26: These settings are used by [`format_guru_cards()` in chainlit.py](https://github.com/navapbc/labs-decision-support-tool/pull/37/files#diff-a69583fbe252b197719af65bbc30c0cc6dfc623f161681b5797e3ad28bd9b4f4R122)

app/src/retrieve.py
```diff
@@ -33,14 +34,31 @@ def retrieve_with_scores(
...
+        ]
+        logger.info("Retrieved %d docs:\n  %s", len(chunks_with_scores), "\n  ".join(retrievals))
 
         # Scores from the DB query are negated, presumably to reverse the default sort order
-        return [ChunkWithScore(chunk, -score) for chunk, score in chunks_with_scores]
+        filtered_chunks_with_scores = [
+            ChunkWithScore(chunk, -score)
+            for chunk, score in chunks_with_scores
+            if -score >= retrieval_k_min_score
+        ]
```
yoomlam, line 56: Logic is changed to use `retrieval_k_min_score`.

app/src/chainlit.py
ccheng26: Do we need this if it's already in the logger?

app/src/chat_engine.py
```diff
@@ -50,9 +61,22 @@ def create_engine(engine_id: str) -> ChatEngineInterface | None:
 class GuruBaseEngine(ChatEngineInterface):
     datasets: list[str] = []
 
+    # Thresholds that determine which documents are sent to the LLM
+    retrieval_k: int = 8
```
ccheng26, line 65: How was the threshold determined?

app/src/format.py
```diff
@@ -1,19 +1,36 @@
...
 from src.db.models.document import ChunkWithScore
 
+logger = logging.getLogger(__name__)
+
 # We need a unique identifier for each accordion,
-# even across multiple calls to this function
-_accordion_id = 0
+# even across multiple calls to this function.
+# Choose a random number to avoid id collisions when hotloading the app during development.
+_accordion_id = random.randint(0, 1000000)
```
ccheng26, line 12: Odds of collisions are low, but should we do a check for if the accordion id exists before assignment as well?

app/src/chainlit.py
yoomlam: It's intended for non-engineers as a reminder of the current setting values, since these settings are newly added. Once the team understands this, we can reduce the verbosity of this.

app/src/chat_engine.py
yoomlam: Arbitrarily chosen: some value greater than the original 5 so that we send more docs to the LLM. Note that this is the default value and the user can adjust it.

ccheng26 (approved):
lgtm

PR #38, opened 2024-07-30
feat: Add LLM selection in Chainlit

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-270

## Changes

- Allow default LLM to be set via env variable.
- Allow the user to select their preferred LLM, by showing a dropdown with the formatted names from `get_models()`.
- When submitting a generation request to the LLM, use the selected LLM model.

## Context for reviewers

Prior to this, the LLM model was hardcoded.

## Testing

1. Start Ollama
2. Add `OLLAMA_HOST=http://host.docker.internal:11434` to `.env` so that there are multiple LLMs to choose from. 
3. Start Chainlit UI, click on the Chainlit settings, select an LLM, submit a request and confirm that the LLM response is generated from the chosen LLM.
   - examine logs to see that the chosen LLM was used, such as `Calling ollama/llama2:latest for query: ...`
image

To set the default LLM, `export LLM='ollama/llama2:latest'` then restart Chainlit.

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -6,7 +6,7 @@ from urllib.parse import parse_qs, urlparse
 import chainlit as cl
 from chainlit.input_widget import InputWidget, Slider
 from src import chat_engine
-from src.app_config import UserConfig, app_config
+from src.app_config import app_config
 from src.format import format_guru_cards
 from src.login import require_login
 
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -4,11 +4,12 @@ from typing import Any
 from urllib.parse import parse_qs, urlparse
 
 import chainlit as cl
-from chainlit.input_widget import InputWidget, Slider
+from chainlit.input_widget import InputWidget, Select, Slider
 from src import chat_engine
 from src.app_config import app_config
 from src.chat_engine import ChatEngineInterface
 from src.format import format_guru_cards
+from src.generate import get_models
 from src.login import require_login
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


app/src/chainlit.py
```diff
@@ -52,6 +53,7 @@ async def _init_chat_settings(engine: ChatEngineInterface) -> dict[str, Any]:
         for setting_name in engine.user_settings
         if setting_name in _WIDGET_FACTORIES
     ]
+    input_widgets.append(_WIDGET_FACTORIES["llm"](app_config.llm or getattr(engine, "llm", None)))
```
yoomlam, line 56: If env var `LLM` is set, use `app_config.llm`, otherwise use the `engine.llm`.

app/src/generate.py
```diff
@@ -66,5 +72,13 @@ def generate(
...
-    response = completion(model="gpt-4o", messages=messages)
+
+    logger.info("Calling %s for query: %s", llm, query)
+    response = completion(model=llm, messages=messages, **completion_args(llm))
     return response["choices"][0]["message"]["content"]
+
+
+def completion_args(llm: str) -> dict[str, Any]:
+    if llm.startswith("ollama/"):
+        return {"api_base": os.environ["OLLAMA_HOST"]}
```
yoomlam, line 83: Apparently, setting `OLLAMA_HOST` isn't sufficient. LiteLLM continues to use the default `localhost` unless `api_base` is passed in as an argument.

app/src/chainlit.py
```diff
@@ -118,7 +126,7 @@ async def on_message(message: cl.Message) -> None:
 
     engine: chat_engine.ChatEngineInterface = cl.user_session.get("chat_engine")
     try:
-        result = engine.on_message(question=message.content)
+        result = await cl.make_async(lambda: engine.on_message(question=message.content))()
```
yoomlam, line 129: Based on https://github.com/Chainlit/chainlit/issues/274#issuecomment-1677920286, this needs to be asynchronous to avoid the frontend `Could not reach the server` alert, which pops up when the frontend times out. The backend continues to work and produces a response, but the response never reaches the UI since the frontend times out.

ccheng26 (approved):

PR #39, opened 2024-07-30
feat: Use URL query params to set initial chat settings

yoomlam:
## Ticket

Satisfies https://navalabs.atlassian.net/browse/DST-284 if the mentioned "features" are user-changeable chat settings.

## Changes

- Parses URL query parameters
- For each of the settings allowed to be changed by users (`engine.user_settings`), set the initial setting value to the value in the query parameter, instead of using the engine's default value for the setting.
- Added `test_chainlit.py`

## Testing

Start Chainlit app, go to `chat/?engine=guru-snap&llm=gpt-4o&retrieval_k=3&someunknownparam=42`.
Expect the llm and retrieval_k chat settings to be set as specified in the URL.
Expect UI and log to show: `Unused URL query parameters: {'someunknownparam': '42'}`

Note that the URL query values are not validated against the UI valid values. This could be considered a feature, for example to test `retrieval_k` beyond the chainlit Widget's range.


app/src/chainlit.py
```diff
@@ -46,12 +59,18 @@ def _init_chat_engine(engine_id: str) -> ChatEngineInterface | None:
...
     input_widgets: list[InputWidget] = [
-        _WIDGET_FACTORIES[setting_name](getattr(engine, setting_name))
+        _WIDGET_FACTORIES[setting_name](
+            query_values.pop(setting_name, None) or getattr(engine, setting_name)
+        )
         for setting_name in engine.user_settings
         if setting_name in _WIDGET_FACTORIES
     ]
+    if query_values:
+        logger.warning("Unused query values: %r", query_values)
```
yoomlam, line 73: Should a warning be reported to the user as a `cl.Message` as well?

app/tests/src/test_chainlit.py
```diff
@@ -0,0 +1,20 @@
...
+    engine_id = query_values.pop("engine")
+    assert engine_id == "guru-snap"
+
+    engine = chat_engine.create_engine(engine_id)
+
+    input_widgets: list[str] = [
+        query_values.pop(setting_name, None) or getattr(engine, setting_name)
+        for setting_name in engine.user_settings
+        if setting_name in chainlit._WIDGET_FACTORIES
+    ]
```
yoomlam, line 16: Replicates logic in `chainlit.py`

app/src/chainlit.py
```diff
@@ -47,16 +70,19 @@ def _init_chat_engine(engine_id: str) -> ChatEngineInterface | None:
...
-async def _init_chat_settings(engine: ChatEngineInterface) -> dict[str, Any]:
+def _init_chat_settings(
+    engine: ChatEngineInterface, query_values: dict[str, str]
+) -> list[InputWidget]:
     input_widgets: list[InputWidget] = [
-        _WIDGET_FACTORIES[setting_name](getattr(engine, setting_name))
+        _WIDGET_FACTORIES[setting_name](
+            query_values.pop(setting_name, None)
+            or getattr(app_config, setting_name, None)
+            or getattr(engine, setting_name)
```
yoomlam, line 80: URL query parameters take precedence over `app_config` constants, which take precedence over engine default settings. This sets the initial value for the setting in the widget, which can subsequently be modified by the user.

KevinJBoyer (approved):
LGTM!

app/src/chainlit.py
```diff
@@ -31,14 +35,33 @@ async def start() -> None:
         ).send()
         return
 
-    settings = await _init_chat_settings(engine)
+    input_widgets = _init_chat_settings(engine, query_values)
+    settings = await cl.ChatSettings(input_widgets).send()
+    logger.info("Initialized settings: %s", pprint.pformat(settings, indent=4))
+    if query_values:
```
KevinJBoyer, line 41: I hadn't seen this pattern with pop and then checking for remaining before reviewing your PRs, it's really slick!

ccheng26 (approved):
Lgtm, can we link the subsequent ticket to validate the url params? Though there's an alternative for selection, there's a strong likelihood of cases where the param given for the llms won't match.

yoomlam:
> Lgtm, can we link the subsequent ticket to validate the url params? Though there's an alternative for selection, there's a strong likelihood of cases where the param given for the llms won't match.

@ccheng26, are you saying we should create a ticket to validate the URL query params? If so, feel free to create it. 
1. I don't know what the UI should look like if the validation fails. Currently, it just shows a message: https://github.com/navapbc/labs-decision-support-tool/pull/39/files#diff-a69583fbe252b197719af65bbc30c0cc6dfc623f161681b5797e3ad28bd9b4f4R46.
2. I'm unsure how often this URL query parameter feature will be used to warrant spending more time on it. I'd rather wait until we get some user or Designer-Researcher feedback.

yoomlam:
> I think it's less likely some of the params would be out of bounds as much as it is the exact spelling/model string for the llms.

Gotcha, if the LLM value isn't in the list of options, then the UI shows blank:
image

Yea, I'm not sure how much validation we'd want to do since for some parameters, it could be considered a feature:
> Note that the URL query values are not validated against the UI valid values. This could be considered a feature, for example to test retrieval_k beyond the chainlit Widget's range.

PR #40, opened 2024-08-01
feat: chunk and store BEM pdf

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-342 

## Testing

1. add BEM policy manual to /app/documents directory
2. run `make ingest-policy-pdfs DATASET_ID=bridges-eligibility-manual BENEFIT_PROGRAM=multi-benefit BENEFIT_REGION=Michigan FILEPATH=/app/documents`
3. data should appear in DB (`SELECT * from document where dataset='bridges-eligibility-manual'`); get the document_id

Screenshot 2024-08-06 at 9 44 25 AM
4. data should appear in DB (`SELECT * from chunk where document_id='{document_id}'`)

Screenshot 2024-08-06 at 9 41 53 AM

## After deployment
[Load data to dev db](https://nava.slack.com/archives/C06DP498D1D/p1721397891286199?thread_ts=1721396512.937399&cid=C06DP498D1D)


app/src/ingest_policy_pdfs.py
```diff
@@ -22,11 +32,93 @@ def _ingest_policy_pdfs(
     embedding_model = app_config.sentence_transformer
     for file in file_list:
         if file.endswith(".pdf"):
+            with smart_open_file(file, "rb") as fin:
+                extract_text_to_fp(fin, output_string, laparams=LAParams(), output_type="text")
```
KevinJBoyer, line 36: Could you use [extract_text](https://pdfminersix.readthedocs.io/en/latest/reference/highlevel.html#extract-text) here so that this becomes `output_string = extract_text(...` (rather than having to declare a global `output_string = StringIO()`

app/src/ingest_policy_pdfs.py
```diff
@@ -22,11 +32,93 @@ def _ingest_policy_pdfs(
...
+                parse_pdf_and_add_to_db(
+                    contents=output_string.getvalue(),
+                    db_session=db_session,
+                    doc_attribs=doc_attribs,
+                    embedding_model=embedding_model,
+                )
+
             logger.info(
                 f"Processing pdf file: {file} at {pdf_file_dir} using {embedding_model}, {db_session}, with {doc_attribs}"
             )
```
KevinJBoyer, line 47: I think we should move this log statement up, so that it happens before the processing code -- that way if an error comes out, that error will happen after this message (not before)

app/src/ingest_policy_pdfs.py
```diff
@@ -22,11 +32,93 @@ def _ingest_policy_pdfs(
...
+        if is_header and start_new_section and body_content != "":
+            current_title = contents
+        else:
+            body_content += f"{contents}\n"
+
+        # If starting a new section and body content is not empty, process and save it
+        if start_new_section and body_content.strip():
+            # Create and add the document to the database
+            document = Document(name=current_title, content=body_content, **doc_attribs)
+            db_session.add(document)
```
KevinJBoyer, line 76: We want one `Document` per PDF file -- the content of the document should be the full text of the PDF. Then we want to split that text into separate chunks

app/src/ingest_policy_pdfs.py
```diff
@@ -22,11 +32,93 @@ def _ingest_policy_pdfs(
...
             logger.info(
                 f"Processing pdf file: {file} at {pdf_file_dir} using {embedding_model}, {db_session}, with {doc_attribs}"
             )
 
 
+def parse_pdf_and_add_to_db(
+    contents: str,
+    db_session: db.Session,
+    doc_attribs: dict[str, str],
+    embedding_model: SentenceTransformer,
```
KevinJBoyer, line 54: Instead of passing db_session and embedding_model as parameters, grab them from `app_config`

app/src/ingest_policy_pdfs.py
KevinJBoyer: In other words, this code should get pulled out of the loop -- a rough structure could be:

```
pdf_text = extract_text(...)
header = get_header(pdf_text)
document = Document(...)
chunks = get_chunks(pdf_text, document)
```

KevinJBoyer (commented):
Nice work on the regex to pull out the document header! I left a comment on code organization -- we only want one Document per PDF file, so in the main loop over the files in the directory, I think it should be something more like:

```
for pdf in directory:
    pdf_text = extract_text(pdf)
    header = get_header(pdf_text)
    document = Document(header, pdf_text, ...)
    chunks = get_chunks(document, pdf_text)

File: app/tests/docs/policy_pdf.pdf
Comment (yoomlam) [inline app/tests/docs/policy_pdf.pdf:1]: Do you want to rename this to `bem_policy.pdf` since it's specific to MI BEM?

File: app/tests/src/test_ingest_policy_pdfs.py [hunk 5]
```diff
@@ -39,8 +35,18 @@ def test__ingest_policy_pdfs(
 
     with caplog.at_level(logging.INFO):
         if file_location == "local":
-            _ingest_policy_pdfs(db_session, policy_local_file, doc_attribs)
+            _ingest_policy_pdfs(policy_local_file, doc_attribs)
```
yoomlam, line 38: Could be simplified to:
```suggestion
            _ingest_policy_pdfs("/app/tests/docs/", doc_attribs)
```

app/src/ingest_policy_pdfs.py
```diff
@@ -14,17 +20,114 @@
...
     embedding_model = app_config.sentence_transformer
+    db_session = app_config.db_session()
     for file in file_list:
         if file.endswith(".pdf"):
-            logger.info(
-                f"Processing pdf file: {file} at {pdf_file_dir} using {embedding_model}, {db_session}, with {doc_attribs}"
+            with smart_open_file(file, "rb") as fin:
+                output_string = extract_text(fin, laparams=LAParams())
+                logger.info(
+                    f"Processing pdf file: {file} at {pdf_file_dir} using {embedding_model}, {db_session}, with {doc_attribs}"
```
yoomlam, line 34: I would put this logger line before `extract_text()` in case there are errors when calling the function or maybe even before `smart_open_file` in case there are problems accessing the file, for example from S3.
What's the usefulness of logging `db_session`? 
I'd log `pdf_file_dir`, `embedding_model`, and `doc_attribs` outside of the `for file` loop since it doesn't change for each file.

app/src/ingest_policy_pdfs.py
```diff
@@ -14,17 +20,114 @@
...
+        if is_header and start_new_section and body_content != "":
+            current_title = contents
+        else:
+            body_content += f"{contents}\n"
+
+    document = Document(name=current_title, content=body_content, **doc_attribs)
+    db_session.add(document)
+
+    process_chunk(body_content, document, db_session)
+    db_session.commit()
```
yoomlam, line 65: Note that `db_session.commit()` is being called once per file. Have you tested this function on multiple files?
I thought `db_session.commit()` should be called once per `db_session`:
```
    with app_config.db_session() as db_session:
        ingestion_call(pdf_file_dir, doc_attribs)
        db_session.commit()
```

yoomlam (approved):
Only major thing is to test ingesting multiple pdf files.

app/src/ingest_policy_pdfs.py
yoomlam: @ccheng26 Did you check if you can call `db_session.commit()` multiple times for the same `db_session`?

yoomlam: Update: multiple `commit()` calls work. we suspect new transactions are created for subsequent `commit()` calls.

PR #42, opened 2024-08-07
fix: Commit after ingesting Guru cards

KevinJBoyer:
## Ticket

n/a

## Changes
 - Pass the DB session as a parameter to the ingest functions

## Context for reviewers
 - The wrapper function in ingest utils controls the commit, so the ingestion function for Guru Cards need to use the same session, otherwise the session it uses never commits

### DIFF
--- a/app/src/util/ingest_utils.py
+++ b/app/src/util/ingest_utils.py
@@ -29,7 +29,7 @@ def process_and_ingest_sys_args(sys: ModuleType, logger: Logger, ingestion_call:
     }
 
     with app_config.db_session() as db_session:
-        ingestion_call(pdf_file_dir, doc_attribs)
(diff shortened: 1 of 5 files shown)


ccheng26 (approved):

app/src/util/ingest_utils.py
```diff
@@ -29,7 +29,7 @@ def process_and_ingest_sys_args(sys: ModuleType, logger: Logger, ingestion_call:
     }
 
     with app_config.db_session() as db_session:
-        ingestion_call(pdf_file_dir, doc_attribs)
+        ingestion_call(db_session, pdf_file_dir, doc_attribs)
         db_session.commit()
```
yoomlam, line 33: Oof! yea, the `db_session` created here isn't the same one used within `ingestion_call()`

PR #44, opened 2024-08-08
feat: Make ingestion drop dataset if already exists

KevinJBoyer:
## Ticket

n/a

## Changes
 - Make `process_and_ingest_sys_args` drop dataset if exists
 - Add coverage for ingest_utils

## Context for reviewers
 - We anticipate reloading data multiple times as we iterate on the ingestion code, so make dropping an existing dataset the default behavior

## Testing
`make test args="tests/src/test_ingest_utils.py`


app/tests/src/test_ingest_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+from src.db.models.document import Document
+from src.util.ingest_utils import _drop_existing_dataset, process_and_ingest_sys_args
+from tests.src.db.models.factories import DocumentFactory
+
+
+def test__drop_existing_dataset(db_session, enable_factory_create):
+    db_session.execute(delete(Document))
+
+    docs = DocumentFactory.create_batch(2)
```
yoomlam, line 14: If there's a miniscule chance that the Documents are given the same dataset string, then to eliminate the risk of a flaky test, create the 2 documents with different `dataset` values.

app/tests/src/test_ingest_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    # This shouldn't do anything
+    assert _drop_existing_dataset(db_session, "nonexistent dataset") is False
+    assert len(db_session.execute(select(Document)).all()) == 2
+
+    # After this, the only document left should be the second one
+    assert _drop_existing_dataset(db_session, docs[0].dataset) is True
+    assert db_session.execute(select(Document)).one()[0].dataset == docs[1].dataset
+
+
+def test_process_and_ingest_sys_args_requires_four_args(caplog):
```
yoomlam, line 25: Nice! Thank you!

app/tests/src/test_ingest_utils.py
```diff
@@ -0,0 +1,115 @@
...
+                "bridges-eligibility-manual",
+                "SNAP",
+                "Michigan",
+                "/some/folder",
+            ],
+            logger,
+            ingest,
+        )
+        assert "Finished processing" in caplog.text
+        ingest.assert_called_with(
```
yoomlam, line 59: TIL: `assert_called_with`

yoomlam (approved):

PR #45, opened 2024-08-08
fix: Generate embeddings based on chunk content

KevinJBoyer:
## Ticket

n/a

## Changes
 - Generate embedding based on chunk content
 - Add test coverage (and simplify existing test)
 - Update the test PDF to include two pages, so that there are multiple chunks generated for the test code

## Context for reviewers
- Previously each chunk was getting an embedding based on the contents of the entire document

## Testing
`make test args="tests/src/test_ingest_policy_pdfs.py`


app/src/ingest_policy_pdfs.py
```diff
@@ -106,11 +104,13 @@ def process_chunk(text: str, document: Document, db_session: db.Session) -> None
             current_chunk.append(sentence)
```
yoomlam, line 104: I think the logic in lines 101-103 might be incorrect. Should it be this?:
```
        total_token_count = len(embedding_model.tokenizer.tokenize("".join(current_chunk)+sentence))

        if total_token_count <= embedding_model.max_seq_length:
```

Also to confirm: we're comparing "tokens" to "seq_length" -- are they comparable?

app/src/ingest_policy_pdfs.py
```diff
@@ -106,11 +104,13 @@ def process_chunk(text: str, document: Document, db_session: db.Session) -> None
...
             current_token_count += token_count
         else:
+            chunk_text = "".join(current_chunk)
+            chunk_embedding = embedding_model.encode(chunk_text, show_progress_bar=False)
```
yoomlam, line 108: Note that `encode()` accepts a `List[str]`, so we may not need to `"".join(current_chunk)`.

KevinJBoyer: That encodes each str separately and then returns a list of embeddings for each: https://sbert.net/docs/package_reference/sentence_transformer/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode

app/src/ingest_policy_pdfs.py
yoomlam: I'm thinking they are not comparable.
* a token is a string, typically a word or part of a word
* `max_seq_length` refers to the length of a vector of floats, where each float could reflect a word, several words, or a concept.

KevinJBoyer: Sorry, didn't see your comment as I was writing mine

> max_seq_length refers to the length of a vector of floats, where each float could reflect a word, several words, or a concept.

Where are you seeing that? https://sbert.net/docs/package_reference/sentence_transformer/SentenceTransformer.html#sentence_transformers.SentenceTransformer.get_max_seq_length

yoomlam: Okay, if they're comparable, then the code is equivalent.

yoomlam: > Where are you seeing that?

Just using my prior NLP experience. I'll dig into their docs to find a relationship between `tokenize()` and `encode()`.
Consider this comment nonblocking.

yoomlam (approved):

app/src/ingest_policy_pdfs.py
KevinJBoyer: tokenizer.tokenize gives you the underling tokens:

```
>>> embedding_model.tokenizer.tokenize("Hello world.")
['hello', 'world', '.']
```

encode gives the the actual embedding:

```
>>> embedding_model.encode("Hello world.")
array([ 6.10581562e-02,  3.09772529e-02,  1.49336702e-03,  1.42844126e-03,
...
```

yoomlam: Yea, to me, their lengths are not comparable, i.e, a float does not necessarily map one-to-one to a token.

KevinJBoyer: max_seq_length is the maximum number of tokens that can be used to generate the embedding. The size of the embedding is unrelated. E.g., for mpnet, the maximum number of tokens that can be provided as input to the embedding model is 512, while the size of the embeddings that mpnet produces is 768

yoomlam: Ah, okay. My interpretation of `max_seq_length` is incorrect then. I thought it was characterizing the output embedding size.

yoomlam: Totally misread this:
> Returns the maximal sequence length that the model _accepts_. Longer _inputs_ will be truncated.

KevinJBoyer: Yeah, max_seq_length is not the name I would have picked 😅

PR #49, opened 2024-08-13
fix: Read titles from PDF metadata

KevinJBoyer:
## Ticket

n/a

## Changes
 - Read Document.name from BEM policy PDFs from the PDF metadata, and the BEM number from the filename.

## Context for reviewers
 - This fixes a bug that was causing BEMs with letters after the number to not be given a number

## Testing
Before patch:
Screenshot 2024-08-13 at 9 16 35 AM 1

After patch and re-ingesting:
image


app/src/ingest_policy_pdfs.py
```diff
@@ -28,12 +42,12 @@ def _ingest_policy_pdfs(
     embedding_model = app_config.sentence_transformer
 
     logger.info(f"Processing pdfs {pdf_file_dir} using {embedding_model} with {doc_attribs}")
```
KevinJBoyer, line 44: Minor refactors here: rename file -> file_path, and fin -> file. Since we're not setting any LAParams, we can let pdfminer use its defaults rather than pass it ourselves.

app/src/ingest_policy_pdfs.py
```diff
@@ -45,20 +59,17 @@ def parse_pdf_and_add_to_db(
...
         )
         # Check if we need to start a new section
-        if is_header and start_new_section and body_content != "":
-            current_title = contents
-        else:
+        if not (is_header and start_new_section and body_content != ""):
             body_content += f"{contents}\n"
 
-    document = Document(name=current_title, content=body_content, **doc_attribs)
+    document = Document(content=body_content, **doc_attribs)
```
KevinJBoyer, line 72: Code changes above reflect that current_title is no longer used.

app/src/ingest_policy_pdfs.py
```diff
@@ -45,20 +59,17 @@ def parse_pdf_and_add_to_db(
...
     start_new_section = True
     for text_contents in text_split_by_header:
         is_header, contents, start_new_section = get_header_and_is_current_section(
             text_contents, start_new_section
         )
         # Check if we need to start a new section
-        if is_header and start_new_section and body_content != "":
-            current_title = contents
-        else:
+        if not (is_header and start_new_section and body_content != ""):
```
yoomlam, line 69: This is easier to interpret for me:
```suggestion
        if not is_header or not start_new_section or not body_content:
```

yoomlam (approved):

app/src/ingest_policy_pdfs.py
```diff
@@ -45,20 +59,17 @@ def parse_pdf_and_add_to_db(
     # Match header in BEM manual
     header_pattern = r"(BEM\s\d*\s+\d+\sof\s\d+\s+\w.*)"
```
yoomlam, line 60: Just noticed: `header_pattern` can be refactored into a constant.

app/src/ingest_policy_pdfs.py
```diff
@@ -45,20 +59,17 @@ def parse_pdf_and_add_to_db(
...
     text_split_by_header = re.split(header_pattern, contents)
-    current_title = ""
     body_content = ""
     start_new_section = True
```
yoomlam, line 63: Line 63 isn't needed
```suggestion
```

PR #50, opened 2024-08-13
feature: Log metadata for retrieved chunks

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-373

## Changes
 - Log each retrieved chunk's document name, ID, and similarity score to the query in the metadata for Chainlit

## Context for reviewers
 - n/a

## Testing
Removing the cast to str for settings is safe:
image

Metadata in LiteralAI dashboard:
image

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -1,6 +1,6 @@
 import logging
 import pprint
-from typing import Any
+from typing import Any, Sequence
 from urllib.parse import parse_qs, urlparse
 
 import chainlit as cl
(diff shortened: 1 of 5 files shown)


app/tests/src/test_format.py
```diff
@@ -42,26 +42,18 @@ def test_format_guru_cards_with_score(monkeypatch, app_config, db_session, enabl
     assert len(_unique_accordion_ids(html + next_html)) == 2 * len(chunks_with_scores)
 
 
-def _chunks_with_scores():
```
KevinJBoyer, line 45: Pulled this into a fixture, refactored test_format and test_generate to use it.

app/src/chainlit.py
```diff
@@ -47,7 +48,7 @@ async def start() -> None:
 
     await cl.Message(
         author="backend",
-        metadata={"engine": engine_id, "settings": str(settings)},
+        metadata={"engine": engine_id, "settings": settings},
```
KevinJBoyer, line 51: metadata can contain json, so removed this cast.

yoomlam (approved):

PR #52, opened 2024-08-15
feat: Link to BEM documents

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-375

## Changes
 - Link accordion titles to BEM document
 - Link references in chunk text

## Context for reviewers
 - I'm a little torn on how to handle the case of no matching BEM ### in `_get_bem_url` -- raising an exception felt a bit cleaner, but an alternative would be to return `str | None`.

## Testing
Screenshot 2024-08-15 at 8 46 50 AM
Screenshot 2024-08-15 at 8 46 57 AM

### DIFF
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -12,6 +12,9 @@ logger = logging.getLogger(__name__)
 # Choose a random number to avoid id collisions when hotloading the app during development.
 _accordion_id = random.randint(0, 1000000)
 
+# Regular expression to match BEM followed by 3 digits, optionally followed by a letter
+BEM_PATTERN = r"(BEM\s(\d{3}[A-Z]?))"
+
 
 def format_guru_cards(
     chunks_shown_max_num: int,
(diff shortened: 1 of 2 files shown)


app/tests/src/test_format.py
```diff
@@ -71,6 +77,8 @@ def test__format_to_accordion_html(app_config, db_session, enable_factory_create
 
 def test_format_bem_documents():
     docs = DocumentFactory.build_batch(4)
+    for doc in docs:
+        doc.name += "BEM 123"
```
KevinJBoyer, line 81: Ensure that the doc names have a valid BEM number for link replacement by the formatter

app/src/format.py
```diff
@@ -119,12 +124,27 @@ def _format_to_accordion_group_html(documents: OrderedDict[Document, list[ChunkW
...
-            </div>"""
+            </div>"""  # noqa: B907
 
     return "<h3>Source(s)</h3>" + html if html else ""
+
+
+def _get_bem_url(text: str) -> str:
+    bem = re.search(BEM_PATTERN, text)
+    if not bem:
+        raise ValueError(f"No BEM number found in text: {text}")
```
yoomlam, line 141: Missing unit test for error.

yoomlam (approved):

app/src/format.py
KevinJBoyer: Good call, added!

PR #56, opened 2024-08-23
feat: Group markdown text list items

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-416

## Changes
 - Add several dataclasses
 - Add function to return grouped texts -- concatenating list items together, changing the overall `TextType` to `List`, and logging a warning if the headers don't match

## Context for reviewers
 - Does putting these in a new folder `ingestion` make sense? Should we call this `bem-ingestion`?
 - Open to feedback on the data structures. Do we need a separate Page to capture the difference between the page index in the document, and the page's human label?

## Testing
`make test args=tests/src/ingestion/test_get_grouped_texts.py`


app/src/ingestion/elements.py
```diff
@@ -0,0 +1,34 @@
...
+
+@dataclass
+class Heading:
+    title: str
+    level: int
+    page: Page
+
+
+@dataclass
+class EnrichedText:
```
KevinJBoyer, line 30: It felt a bit like overdoing it to add a new class `GroupedText` with the exact same fields, but open to the team's thoughts here!

app/src/ingestion/elements.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@dataclass
+class Page:
+    # The actual page in the document
+    pdf_page_number: int
+
+    # The label of the page in the document
+    # E.g., "i", "ii", "1", "2", etc.
+    document_page_number: str
```
KevinJBoyer, line 19: Wonder if we'd be OK just using an `int` for the index of the page in the document rather than this class. Also I made some slight renames here for clarity but don't feel strongly about this if folks have alternative style preferences

ccheng26: did we want to store the total amount of pages somewhere?

ccheng26 (approved):
left a question otherwise, lgtm

app/src/ingestion/elements.py
yoomlam: > did we want to store the total amount of pages somewhere?

That would go in a PdfInfo or DocInfo class, if we needed it.

app/src/ingestion/elements.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@dataclass
+class Heading:
+    title: str
+    level: int
+    page: Page
```
yoomlam, line 26: I only use an `int` page number: https://github.com/navapbc/labs-decision-support-tool/pull/57/files#diff-aecde49f56fa9585be76a3e8af06031a4037575977c2030c02052a7a9af2c34eR16

Having to populate `Page.document_page_number` involves parsing the pdf text.

We'll have to resolve the 2 `Heading` classes.

app/src/ingestion/elements.py
yoomlam: We don't need to create another class. The tech spec uses different names to distinguish the differences of the I/O for each step to avoid confusion.

app/src/ingestion/elements.py
KevinJBoyer: Left a related comment here -- https://github.com/navapbc/labs-decision-support-tool/pull/57#discussion_r1729203838 I'll follow your lead here, let me know how you'd like me to sync these up and I'll make the change in this PR before merging.

app/src/ingestion/elements.py
yoomlam, line 1: `elements` sounds vague. Can we be more specific to convey the purpose, i.e., what should belong and not belong in this file?

KevinJBoyer: Agree. Maybe pdf_elements? document_elements?

yoomlam: Let's go with the more specific one: `pdf_elements`

app/src/ingestion/pdf_elements.py
```diff
@@ -0,0 +1,24 @@
...
+    NARRATIVE_TEXT = "NarrativeText"
+    LIST_ITEM = "ListItem"
+    LIST = "List"
+
+
+@dataclass
+class Heading:
+    title: str
+    level: int
+    page_number: int | None = None
```
yoomlam, line 16: Should my PR be using this Heading?

app/src/ingestion/get_grouped_texts.py
```diff
@@ -0,0 +1,35 @@
...
+        previous_text = grouped_texts[-1]
+
+        if current_text.type == TextType.LIST_ITEM:
+            # Concatenate the current text to the previous one
+            previous_text.text += "\n" + current_text.text
+            previous_text.type = TextType.LIST
+
+            # Headings should match for list items
+            if current_text.headings != previous_text.headings:
+                logger.warning("Warning: Headings don't match for list items.")
```
yoomlam, line 29: Add some detail to identify the texts.
```suggestion
                logger.warning("Warning: Headings don't match for list items: %s %s", previous_text.text, current_text.text)
```

app/tests/src/ingestion/test_get_grouped_texts.py
```diff
@@ -0,0 +1,75 @@
...
+            text="Narrative starting a new list.",
+            type=TextType.NARRATIVE_TEXT,
+            headings=[Heading(title="Overview", level=1)],
+        ),
+        EnrichedText(
+            text="First item in new list.",
+            type=TextType.LIST_ITEM,
+            headings=[Heading(title="Overview", level=1)],
+        ),
+    ]
```
yoomlam, line 54: Note: I could see reusing this as a factory or fixture for other tests.

app/src/ingestion/get_grouped_texts.py
```diff
@@ -0,0 +1,35 @@
...
+
+    if not markdown_texts:
+        return []
+
+    grouped_texts = [markdown_texts[0]]
+
+    for current_text in markdown_texts[1:]:
+        previous_text = grouped_texts[-1]
+
+        if current_text.type == TextType.LIST_ITEM:
```
yoomlam, line 22: Need to check if `previous_text.text.rstrip()` ends in `:`.

Requirement 4 states:
> For list items: Keep grouped list items, nested list items, + the preceding sentence (which should end in a : ) together

Also
> [TASK 3.a] Split overly long text in grouped_texts (i.e., which don’t fit within the embedding model limit) into splits (a list of strings)
> - If a list is split, prepend the list-preceding sentence (ending with :) to the beginning of each splits[i].

PR #57, opened 2024-08-23
feat: Add extract_outline() PDF utility

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-413

## Changes

Add extract_outline() to extract the heading hierarchy from a PDF's catalog dictionary entry with key "Outlines".

## Testing

Added unit test using one of the BEM pdfs, a small one.


app/src/util/pdf_utils.py
```diff
@@ -0,0 +1,60 @@
...
+            pageno = None
+            if dest:
+                dest = resolve_dest(dest)
+                pageno = pages[dest[0].objid]
+            elif action and isinstance(action, dict):
+                subtype = action.get("S")
+                if subtype and repr(subtype) == "/'GoTo'" and action.get("D"):
+                    dest = resolve_dest(action["D"])
+                    pageno = pages[dest[0].objid]
+            outline.append(Heading(title, level, pageno))
```
KevinJBoyer, line 57: You can improve the clarify of the code here by creating a function to encapsulate the functionality of finding the page number, so this becomes:
```
def get_page_number(dest, action) -> int | None:
    # Logic to parse dest and action into the page number here
    # i.e., the code that currently lives in the for loop and in resolve_dest
    # Since this code is PDFMiner.six specific, some comments may help the reader understand

for level, title, dest, action, _se in outlines:
   pageno = get_page_number(dest, action)
   outline.append(Heading(title, level, pageno))
```

app/src/util/pdf_utils.py
```diff
@@ -0,0 +1,60 @@
...
+from pdfminer.psparser import PSLiteral
+
+
+@dataclass
+class Heading:
+    title: str
+    # Heading level starting from 1
+    level: int
+    # Page number where the heading first appears
+    pageno: int | None
```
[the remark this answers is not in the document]
KevinJBoyer: Alternatively, since we don't yet have a use case for document_page_number, I can also revise my PR to remove the Page class and instead have just an int -- I slightly prefer this option, but LMK either way and I'll update my PR accordingly before merging!

KevinJBoyer (commented):
Nice! Yeah, I think adding a small PDF and adding that as a test case makes sense!

app/src/util/pdf_utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+def extract_outline(doc: PDFDocument) -> list[Heading]:
+    """
+    Adapted from dumppdf.py:dumpoutline()
+    Extracts the heading hierarchy from a PDF's catalog dictionary entry with key "Outlines".
+
+    Usage:
+        with open("100.pdf", "rb") as fp:
+            doc = PDFDocument(PDFParser(fp))
+            outline = extract_outline(doc)
```
KevinJBoyer, line 27: What do you think about changing the signature to take a filepointer and moving `doc = PDFDocument(PDFParser(fp))` inside the code here so this becomes `outline=extract_outline(fp)`? The benefit is that callers of this utility don't need to know anything about the internal functionality relying on pdfminer.six.

app/src/util/pdf_utils.py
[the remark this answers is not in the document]
yoomlam: Sounds good for both comments. We start with `int` and create a `Page` class when needed.

app/src/util/pdf_utils.py
yoomlam: Will do.

app/src/util/pdf_utils.py
yoomlam: This was all copied from `pdfminer.six` so I wanted to keep it comparable it case it gets updated. With that said, I'll make it more readable.

app/src/util/pdf_utils.py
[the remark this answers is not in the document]
KevinJBoyer: Will make this change in mine now!

app/tests/src/util/test_pdf_utils.py
```diff
@@ -0,0 +1,25 @@
+from src.util.pdf_utils import Heading, extract_outline
+
+
+def test_extract_outline():
+    # Create a PDF with a heading hierarchy
+    with open("/app/tests/src/util/707.pdf", "rb") as fp:
```
yoomlam, line 6: @KevinJBoyer I moved `707.pdf` from `docs/` to this folder b/c of:
```
FAILED tests/src/test_ingest_policy_pdfs.py::test__ingest_policy_pdfs[local] - sqlalchemy.exc.MultipleResultsFound: Multiple rows were found when exactly one was required
```

KevinJBoyer (approved):
LGTM!

app/src/util/pdf_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+    pages = {
+        page.pageid: pageno for (pageno, page) in enumerate(PDFPage.create_pages(doc), start=1)
+    }
+
+    def resolve_page_number(dest: Any, action: Any) -> int | None:
+        if dest:
+            dest = resolve_dest(dest)
+            return pages[dest[0].objid]
+        elif action and isinstance(action, dict):
```
KevinJBoyer, line 54: incredibly minor not: changing this to `if` can help clarify that the above if condition returns early

PR #59, opened 2024-08-23
feat: BEM ingest skeleton

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-417

## Changes

- Added `app/src/ingest_bem_pdfs.py`
- Added `EnrichedText.id` field
- Added placeholder unit tests
- Renamed `get_grouped_texts.py` to more general `[pdf_postprocess.py](https://github.com/navapbc/labs-decision-support-tool/pull/59/commits/0df21a0130c47e1314e46202e91fc651fb03db5a)` so we can add other post-processing functions to it, like in #60 

## Testing

Minimal. Added placeholder `TODO`s


app/src/util/pdf_utils.py
```diff
@@ -9,6 +8,41 @@
...
+def as_pdf_doc(pdf: BinaryIO | PDFDocument) -> PDFDocument:
+    if isinstance(pdf, PDFDocument):
+        return pdf
+    else:
+        return PDFDocument(PDFParser(pdf))
+
+
+def get_pdf_info(pdf: BinaryIO | PDFDocument, count_pages: bool = False) -> dict[str, Any]:
+    doc = as_pdf_doc(pdf)
+    assert len(doc.info) == 1
```
KevinJBoyer, line 20: Is there a reason to throw an exception if this fails?

app/src/util/pdf_utils.py
```diff
@@ -9,6 +8,41 @@
...
 
 
+def as_pdf_doc(pdf: BinaryIO | PDFDocument) -> PDFDocument:
+    if isinstance(pdf, PDFDocument):
+        return pdf
+    else:
+        return PDFDocument(PDFParser(pdf))
+
+
+def get_pdf_info(pdf: BinaryIO | PDFDocument, count_pages: bool = False) -> dict[str, Any]:
```
KevinJBoyer, line 18: Could we introduce a dataclass instead of `dict[str, Any]` so the typing is safer?

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+    chunk = Chunk(
+        document=document,
+        content=chunk_text,
+        tokens=current_token_count,
+        mpnet_embedding=chunk_embedding,
+    )
+    db_session.add(chunk)
+
+
+def main() -> None:
```
KevinJBoyer, line 170: You'll want to also add this script to pyproject.toml and app/Makefile so it can be run from the command line for local development, like this: https://github.com/navapbc/labs-decision-support-tool/pull/31/files#diff-0c6c7defde945725f3d878fe20d0866dc29520f5f3816ff91ca56df2211e5451

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+) -> None:
+    file_list = get_files(pdf_file_dir)
+    embedding_model = app_config.sentence_transformer
+
+    logger.info(
+        "Processing PDFs in %s using %s with %s", pdf_file_dir, embedding_model, doc_attribs
+    )
+    for file_path in file_list:
+        if not file_path.endswith(".pdf"):
+            return
```
KevinJBoyer, line 53:
```suggestion
            continue
```

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def _get_bem_title(file: BinaryIO, file_path: str) -> str:
+    """
+    Get the BEM number from the file path (e.g., 100.pdf) and the
+    document title from the PDF meta data, then put the document
+    title in title case (e.g., INTRODUCTION EXAMPLE -> Introduction Example)
+    and combine: "BEM 100: Introduction Example"
+    """
+    pdf_title = pdf_utils.get_pdf_info(file)["title"]
```
KevinJBoyer, line 35: This will throw an exception and interrupt ingestion for the directory if a PDF doesn't have a title:

```suggestion
    pdf_info = pdf_utils.get_pdf_info(file)
    if 'title' not in pdf_info:
        ... # handle this case
    pdf_title = pdf_info["title"]
```

I have a suggestion later to make this function return a dataclass, which enables mypy to catch these

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+from src.util.ingest_utils import process_and_ingest_sys_args
+from src.util.pdf_utils import Heading
+
+logger = logging.getLogger(__name__)
+
+# Print INFO messages since this is often run from the terminal
+# during local development
+logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
+
+HEADER_PATTERN = r"(BEM\s\d*\s+\d+\sof\s\d+\s+\w.*)"
```
KevinJBoyer, line 25: Is this being used?

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+    db_session: db.Session,
+    pdf_file_dir: str,
+    doc_attribs: dict[str, str],
+) -> None:
+    file_list = get_files(pdf_file_dir)
+    embedding_model = app_config.sentence_transformer
+
+    logger.info(
+        "Processing PDFs in %s using %s with %s", pdf_file_dir, embedding_model, doc_attribs
+    )
```
KevinJBoyer, line 50: nit: f-strings are easier to follow when reading code, e.g.:
```suggestion
    logger.info(
        f"Processing PDFs in {pdf_file_dir} using {embedding_model} with {doc_attribs}"
    )
```

KevinJBoyer: Also, I think it would be better to print `app_config.embedding_model` (a string) than `app_config.sentence_transformer` (a class) here (I realize the original policy PDF code does this too)

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+    for file_path in file_list:
+        if not file_path.endswith(".pdf"):
+            return
+
+        logger.info("Processing file: %s", file_path)
+        with smart_open(file_path, "rb") as file:
+            grouped_texts = _parse_pdf(file)
+
+            doc_attribs["name"] = _get_bem_title(file, file_path)
+            document = Document(content=str(grouped_texts), **doc_attribs)
```
KevinJBoyer, line 60: `content` doesn't seem correct to me, I think we want this to be something like:
```
content="\n".join(g.text for g in grouped_texts)
```

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def to_markdown_texts(enriched_texts: list[EnrichedText]) -> list[EnrichedText]:
+    "Placeholder function. Will be implemented for DST-415, probably in a different file."
+    return enriched_texts
+
+
+@dataclass
+class ChunkTemp:
+    "Temporary class as we work through how we want to update document.py:Chunk"
```
KevinJBoyer, line 94: 👍 Nice, this is a clever strategy for preserving flexibility

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+
+def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
+    enriched_texts = enrich_texts(file)
+    markdown_texts = to_markdown_texts(enriched_texts)
+    grouped_texts = group_texts(markdown_texts)
+
+    # Assign unique ids to each grouped text before they get split into chunks
+    for text in grouped_texts:
+        text.id = str(uuid.uuid1())
+    assert len(set(text.id for text in grouped_texts)) == len(grouped_texts)
```
KevinJBoyer, line 75: FWIW I don't think this is necessary -- I don't think a collision is possible for UUID1, and while it's theoretically possible for UUID4, my go-to reference for UUID4 is that it's more likely a meteor takes out the server the code is running on first

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,171 @@
...
+    "Placeholder function. Will be implemented for DST-414, probably in a different file."
+    outline: list[Heading] = pdf_utils.extract_outline(file)
+    for _heading in outline:
+        pass
+    return []
+
+
+def to_markdown_texts(enriched_texts: list[EnrichedText]) -> list[EnrichedText]:
+    "Placeholder function. Will be implemented for DST-415, probably in a different file."
+    return enriched_texts
```
KevinJBoyer, line 89: Maybe `add_markdown` (for consistency with the verb + noun naming of enrich_texts and group_texts)?

KevinJBoyer (commented):
LGTM overall, thanks for setting this up so that we have a framework to build on in subsequent PRs -- left some comments, additionally, let's add test coverage to the functions in pdf_utils.py.

app/src/ingest_bem_pdfs.py
yoomlam: Nope!

app/src/ingest_bem_pdfs.py
yoomlam: Oops!

app/src/ingest_bem_pdfs.py
yoomlam: I'll remove it if you are sure. Take a gander at https://stackoverflow.com/questions/23983079/using-pythons-uuid-to-generate-unique-ids-should-i-still-check-for-duplicates.

app/src/ingest_bem_pdfs.py
yoomlam: Added this task to https://navalabs.atlassian.net/browse/DST-419

app/src/util/pdf_utils.py
yoomlam: Sure!

app/src/util/pdf_utils.py
yoomlam: The assertion should throw an AssertionError if it's not true. I'll add an error message.

app/src/ingest_bem_pdfs.py
yoomlam: Agree, but Pylint suggests not using f-strings with loggers:
image

KevinJBoyer (approved):
LGTM!

app/src/ingest_bem_pdfs.py
KevinJBoyer: I get where Pylint is coming from, but in this case we're always executing this logging statement (and none of the interpolations are expensive) so there's no harm besides maybe setting a misleading precedent. I don't feel strongly either way!

app/src/ingest_bem_pdfs.py
```diff
@@ -0,0 +1,172 @@
...
+    split_index: int = 0
+
+
+def split_into_chunks(document: Document, grouped_texts: list[EnrichedText]) -> list[ChunkTemp]:
+    """
+    Given EnrichedTexts, convert the text to chunks and add them to the database.
+    """
+    chunks: list[ChunkTemp] = []
+    for paragraph in grouped_texts:
+        assert paragraph.id is not None
```
ccheng26, line 129: should we be handling the exception or will this be implemented in another PR?

yoomlam: This assertion shouldn't fail unless there's a bug since `id` should be populated at this point in the code. It was added to appease MyPy:
image

PR #61, opened 2024-08-24
fix: Merge list items only if they have the same heading

yoomlam:
## Changes

Merge list items only if they have the same heading, instead of logging a warning when they have different headings.

## Testing

Updated test_concatenate_list_items() to include LIST_ITEM in new heading that should not be merged.

### DIFF
--- a/app/tests/src/ingestion/test_get_grouped_texts.py
+++ b/app/tests/src/ingestion/test_get_grouped_texts.py
@@ -56,6 +56,11 @@ def test_concatenate_list_items():
             type=TextType.LIST_ITEM,
             headings=[Heading(title="Overview", level=1)],
         ),
+        EnrichedText(
+            text="New list item in new section.",
+            type=TextType.LIST_ITEM,
+            headings=[Heading(title="Section 1", level=1)],
+        ),
     ]
 
     result = get_grouped_texts(texts)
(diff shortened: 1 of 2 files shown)


app/tests/src/ingestion/test_get_grouped_texts.py
```diff
@@ -81,4 +86,9 @@ def test_concatenate_list_items():
             type=TextType.LIST,
             headings=[Heading(title="Overview", level=1)],
         ),
+        EnrichedText(
+            text="New list item in new section.",
+            type=TextType.LIST_ITEM,
+            headings=[Heading(title="Section 1", level=1)],
+        ),
```
KevinJBoyer, line 93: Commenting here because it's a good illustration: if there's a single list item in the parsed PDF (i.e., it isn't merged with an adjacent paragraph or list item), do we want the type to be LIST_ITEM or LIST? Currently get_grouped_texts always outputs NARRATIVE_TEXT or LIST, this PR also introduces LIST_ITEM. Not a big deal either way, but worth considering which will make future code simpler (since later code that consumes the output from get_grouped_texts and wants to inspect text.type now has to check for both LIST and LIST_ITEM)

KevinJBoyer (approved):
LGTM!

PR #66, opened 2024-08-28
feat: Merges texts that are split across consecutive pages

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-425

## Changes

Added `_should_merge_text()` that merges texts (`NarrativeText` or `List`) that are split across consecutive pages

## Testing

Updated test for `group_texts()`

### DIFF
--- a/app/src/ingestion/pdf_postprocess.py
+++ b/app/src/ingestion/pdf_postprocess.py
@@ -56,6 +56,7 @@ def _apply_stylings(e_text: EnrichedText) -> EnrichedText:
         e_text.stylings = None
     else:
         e_text.stylings = [s for s in e_text.stylings if s not in applied]
+        logger.warning("Stylings were not applied: %s", e_text.stylings, extra={"e_text": e_text})
     return e_text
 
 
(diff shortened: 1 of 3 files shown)


app/tests/src/ingestion/test_pdf_postprocess.py
```diff
@@ -86,41 +86,55 @@ def test_concatenate_list_items():
...
         EnrichedText(
-            text="New list item in new section.",
+            text="New list item in new section",
             type=TextType.LIST_ITEM,
             headings=[Heading(title="Section 1", level=1)],
+            page_number=1,
+        ),
+        EnrichedText(
+            text="with continuing sentence on next page",
+            type=TextType.NARRATIVE_TEXT,
```
yoomlam, line 135: I don't know if unstructured would categorize a list item that overflows onto the next page as `NARRATIVE_TEXT` or `LIST_ITEM`.
If it was `LIST_ITEM`, then `_group_list_texts()` would make it a new item, which is not the correct semantic.

app/src/ingestion/pdf_postprocess.py
```diff
@@ -137,9 +137,9 @@ def _should_merge_list_text(text: EnrichedText, next_text: EnrichedText) -> bool
     return text.type == TextType.NARRATIVE_TEXT and text.text.rstrip().endswith(":")
 
 
-def group_texts(markdown_texts: list[EnrichedText]) -> list[EnrichedText]:
+def _group_list_texts(markdown_texts: list[EnrichedText]) -> list[EnrichedText]:
     """
-    Given EnrichedTexts, concatenate list tems together
+    Given EnrichedTexts, concatenate list items together
```
KevinJBoyer, line 142: Nice catch

app/src/ingestion/pdf_postprocess.py
```diff
@@ -160,3 +160,35 @@ def group_texts(markdown_texts: list[EnrichedText]) -> list[EnrichedText]:
...
     return grouped_texts
+
+
+def _should_merge_text(text: EnrichedText, next_text: EnrichedText) -> bool:
+    "Merges texts that are split across consecutive pages"
+    if text.headings != next_text.headings:
+        return False
+
+    assert text.page_number is not None
+    assert next_text.page_number is not None
```
KevinJBoyer, line 171: We could also log an error and then return False here -- defer to you on the desired behavior: bail out of the whole ETL pipeline (fail an assert) or skip this text but log that something unexpected happened

app/tests/src/ingestion/test_pdf_postprocess.py
KevinJBoyer: My wild guess is that it'll be NarrativeText, with about 50% confidence 😂  I'd say let's try it and see what happens!

KevinJBoyer (approved):
LGTM!

app/src/ingestion/pdf_postprocess.py
yoomlam: At this step in the pipeline, all EnrichedText elements should have page numbers. If not, we should fail and fix the bug.

PR #68, opened 2024-08-29
feat: enrich text using unstructured data

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-414 

## Changes

> added unstructured lib
> added headers to chunk data

## Testing

Run `make ingest-bem-pdfs DATASET_ID=bridges-eligibility-manual BENEFIT_PROGRAM=multiprogram BENEFIT_REGION=Michigan FILEPATH=/app/documents/BEM`

DB should populate with headers
Screenshot 2024-08-29 at 6 33 06 PM


app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +80,59 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+            headings=current_header,
+            id=element._element_id,
+        )
+
+        enrich_text_list.append(enriched_text_item)
+    return enrich_text_list
+
+
+def match_heading(
+    heading_list: list[Heading], heading_name: str, page_number: int | None
```
KevinJBoyer, line 130: What do you think about renaming `heading_list` to `outline` here, to match the name of the variable we pass to this function?

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +80,59 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
-        pass
-    return []
+    current_header = []
+    current_header_level = 1
+    for element in unstructured_json:
+        if element.category == "Footer" or element.category == "Header":
+            continue
+        if element.category == "Title":
+            header = match_heading(outline, element.text, element.metadata.page_number)
+            if header:
```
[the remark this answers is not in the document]
KevinJBoyer: A few other thoughts on this section:
 - I think "heading" is a better name than "header" here (we're trying to use header/fooder exclusively for the top and bottom of pages), so these would be `current_heading`, etc. (Or maybe `current_headings`, since it's a list)
 - could you separate this out into a function, like get_current_heading(header, current_header) 
 - could we treat current_header_level = 0 when current_header is []? If we refactor the code to handle this, then we don't need the current_header_level variable at all, since the current level will always be `len(current_header)`

KevinJBoyer (commented):
Nice work on this! Left some style suggestions for your thoughts. One additional ask, could we add testing of the `enrich_texts` function directly?

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +80,59 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+
+        enrich_text_list.append(enriched_text_item)
+    return enrich_text_list
+
+
+def match_heading(
+    heading_list: list[Heading], heading_name: str, page_number: int | None
+) -> Heading | None:
+    for heading in heading_list:
+        if heading_name.lower() in heading.title.lower() and heading.pageno == page_number:
```
yoomlam, line 133: I learned `casefold()` is better for case-insensitive comparisons: https://stackoverflow.com/questions/45745661/lower-vs-casefold-in-string-matching-and-converting-to-lowercase

See https://github.com/navapbc/labs-decision-support-tool/blob/886b654922b62a7ad64aa11a1b3296dfe528c0f7/app/src/ingestion/pdf_stylings.py#L143

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +80,59 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+        )
+
+        enrich_text_list.append(enriched_text_item)
+    return enrich_text_list
+
+
+def match_heading(
+    heading_list: list[Heading], heading_name: str, page_number: int | None
+) -> Heading | None:
+    for heading in heading_list:
```
yoomlam, line 132: FYI: Instead of iterating through all headings, I perform a stricter heading match algorithm that expects the next heading -- see https://github.com/navapbc/labs-decision-support-tool/blob/886b654922b62a7ad64aa11a1b3296dfe528c0f7/app/src/ingestion/pdf_stylings.py#L139.

This could help identify any mismatch between the outline (`heading_list`) and unstructure's parsed `Title` elements. Also note that it is possible (though unlikely) that the same heading text may appear multiple times in the outline and on the same page.

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +79,62 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+def match_heading(
+    outline: list[Heading], heading_name: str, page_number: int | None
+) -> Heading | None:
+    for heading in outline:
+        if heading_name.lower() in heading.title.lower() and heading.pageno == page_number:
+            return heading
+    return None
+
+
+def _get_current_heading(outline, element, current_headings):
```
yoomlam, line 126:
```suggestion
def _get_current_headings(outline, element, current_headings):
```

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +79,62 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+    if heading := match_heading(outline, element.text, element.metadata.page_number):
+        if heading.level == 1:
+            current_headings = [heading]
+        else:
+            if heading.title != current_headings[-1].title:
+                if len(current_headings) == heading.level:
+                    current_headings = current_headings[:-1]
```
yoomlam, line 133: I don't think this is handling the case where you move from a heading level 3 to a level 1. Line 133 is only removing/popping one level. See https://github.com/navapbc/labs-decision-support-tool/blob/886b654922b62a7ad64aa11a1b3296dfe528c0f7/app/src/ingestion/pdf_stylings.py#L157-L159

ccheng26: this should only be checking the case where the heading levels are the same and replacing the current level. ex: replacing  Heading(title='Family Independence Program (FIP)', level=2, pageno=1) with Heading(title='Program Goal', level=2, pageno=1); I'll add a case to check for the other level jump

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +78,72 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+                if len(current_headings) > heading.level:
+                    current_headings = current_headings[
+                        : -(len(current_headings) - 1 - heading.level)
+                    ]
```
yoomlam, line 142: Add a unit test for this logic. I believe you can then remove lines 137-138 b/c these lines 140-142 should(?) also cover the scenario in lines 137-138.

app/tests/src/test_ingest_bem_pdfs.py
```diff
@@ -55,3 +83,30 @@ def test__ingest_policy_pdfs(caplog, app_config, db_session, policy_s3_file, fil
...
         # assert math.isclose(second_chunk.mpnet_embedding[0], -0.82242084, rel_tol=1e-3)
+
+
+def test__enrich_text():
+    with smart_open("/app/tests/docs/100.pdf", "rb") as file:
+        enriched_text_list = _enrich_texts(file)
+
+        assert len(enriched_text_list) == 9
+        enriched_text_item = enriched_text_list[0]
+        assert isinstance(enriched_text_item, EnrichedText)
```
yoomlam, line 94: Add assertions to check that `EnrichedText` items have the correct headings, type, and page_number associated with them.

yoomlam (changes requested):

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +78,71 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+            )
+    return enrich_text_list
+
+
+def _match_heading(
+    outline: list[Heading], heading_name: str, page_number: int | None
+) -> Heading | None:
+    for heading in outline:
+        if heading.pageno == page_number:
+            # account for spacing differences in unstructured and pdfminer parsing
```
yoomlam, line 125: Oh! Can you provide an example?

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +78,71 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+
+def _get_current_heading(
+    outline: list[Heading], element: Element, current_headings: list[Heading]
+) -> list[Heading]:
+    if heading := _match_heading(outline, element.text, element.metadata.page_number):
+        if heading.level == 1:
+            current_headings = [heading]
+        else:
+            if heading.title != current_headings[-1].title:
+                current_headings = current_headings[: -(len(current_headings) - 1 - heading.level)]
```
yoomlam, line 141: Why isn't this just:
```suggestion
                current_headings = current_headings[:heading.level]
```
i.e., copy all the headings up to the new `heading`.

app/src/ingest_bem_pdfs.py
yoomlam: Is [this](https://github.com/navapbc/labs-decision-support-tool/pull/68/files#diff-ed2bad192257ddbf1660eab275f187aca890343c45cd7131160e9830d773b502R114) the actual output from unstructured?

ccheng26: Interestingly enough the output from unstructured was `TIME AND ATTENDANCE REVIEW PROCESS` (without an extra space) from the pdfminer outline function was `TIME AND ATTENDANCE REVIEW  PROCESS` with an extra space (which is how it looks when reading it)

app/src/ingest_bem_pdfs.py
ccheng26: there's an off by one issue if it's not negative, revised in latest commit

app/src/ingest_bem_pdfs.py
yoomlam: I see https://github.com/navapbc/labs-decision-support-tool/blob/03e45cb09d8f44103e5a6705fc17411641bee9ed/app/tests/src/util/test_pdf_utils.py#L29

image

app/tests/src/test_ingest_bem_pdfs.py
```diff
@@ -55,3 +88,54 @@ def test__ingest_policy_pdfs(caplog, app_config, db_session, policy_s3_file, fil
...
+    replaced_second_level = _get_current_heading(mock_outline, mock_elements[2], mock_outline[:2])
+    assert replaced_second_level == [
+        Heading(title="Overview", level=1, pageno=1),
+        Heading(title="Program Goal", level=2, pageno=1),
+    ]
+
+    dropped_level = _get_current_heading(mock_outline, mock_elements[4], mock_outline)
+    assert dropped_level == [
+        Heading(title="Test Level 1", level=1, pageno=2),
+    ]
```
yoomlam, line 141: This last test isn't executing line 141, which I believe is what you intend.

app/src/ingest_bem_pdfs.py
```diff
@@ -76,15 +78,71 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
...
+
+def _get_current_heading(
+    outline: list[Heading], element: Element, current_headings: list[Heading]
+) -> list[Heading]:
+    if heading := _match_heading(outline, element.text, element.metadata.page_number):
+        if heading.level == 1:
+            current_headings = [heading]
+        else:
+            if heading.title != current_headings[-1].title:
+                current_headings = current_headings[: -(heading.level - 1)]
```
yoomlam, line 141:
```suggestion
                current_headings = current_headings[:(heading.level - 1)]
```

app/tests/src/test_ingest_bem_pdfs.py
yoomlam: Try this:
```suggestion
    current_headings = [
        Heading(title="Overview", level=1, pageno=1),
        Heading(title="Program Goal", level=2, pageno=1),
        Heading(title="Tertiary Program Goal", level=3, pageno=2),
        Heading(title="4th Program Goal", level=4, pageno=2),
    ]
    element = Text(text="Test Level 2", metadata=ElementMetadata(page_number=2))
    dropped_level = _get_current_heading(mock_outline, element, current_headings)
    assert dropped_level == [
        Heading(title="Overview", level=1, pageno=1),
        Heading(title="Test Level 2", level=2, pageno=2),
    ]
```
after updating `mock_outline()` to:
```python
def mock_outline():
    return [
        Heading(title="Overview", level=1, pageno=1),
        Heading(title="Family Independence Program (FIP)", level=2, pageno=1),
        Heading(title="Program Goal", level=2, pageno=1),
        Heading(title="Medical Assistance Program", level=2, pageno=2),
        Heading(title="Program Goal", level=2, pageno=2),
        Heading(title="Tertiary Program Goal", level=3, pageno=2),
        Heading(title="4th Program Goal", level=4, pageno=2),
        Heading(title="Test Level 2", level=2, pageno=2),
    ]

```

app/tests/src/test_ingest_bem_pdfs.py
```diff
@@ -62,12 +100,60 @@ def test__ingest_bem_pdfs(caplog, app_config, db_session, policy_s3_file, file_l
...
+
+    replaced_second_level = _get_current_heading(mock_outline, mock_elements[2], mock_outline[:2])
+    assert replaced_second_level == [
+        Heading(title="Overview", level=1, pageno=1),
+        Heading(title="Program Goal", level=2, pageno=1),
+    ]
+
+    dropped_level = _get_current_heading(mock_outline, mock_elements[4], mock_outline)
+    assert dropped_level == [Heading(title="Test Level 1", level=1, pageno=2)]
```
yoomlam, line 153: Is this `dropped_level` test intended to run [this line](https://github.com/navapbc/labs-decision-support-tool/pull/68/files#diff-d74557876d257737a4a1149fd26cd9a7ef610c44d243ebab30edc8036c65cdeeR144)? If so, it doesn't.

yoomlam: See https://github.com/navapbc/labs-decision-support-tool/pull/68#discussion_r1742539215 for suggestion.

yoomlam (approved):

PR #69, opened 2024-09-03
feat: Save BEM JSON chunks to S3

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-419

## Changes
 - Save chunks as json to S3
 - Minor performance refactor to parallelize generating embeddings for chunk.content
 - Minor fix for a prior DB migration


app/src/db/migrations/versions/2024_09_03_add_num_splits_and_split_index.py
```diff
@@ -0,0 +1,70 @@
...
+revision = "9543d555fad7"
+down_revision = "39d60b8215fa"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("chunk", sa.Column("num_splits", sa.Integer(), nullable=False))
+    op.add_column("chunk", sa.Column("split_index", sa.Integer(), nullable=False))
```
KevinJBoyer, line 21: These were missed in a prior migration (see the code changes in document.py)

app/src/ingest_bem_pdfs.py
```diff
@@ -26,9 +27,8 @@
 def _get_bem_title(file: BinaryIO, file_path: str) -> str:
     """
     Get the BEM number from the file path (e.g., 100.pdf) and the
-    document title from the PDF meta data, then put the document
-    title in title case (e.g., INTRODUCTION EXAMPLE -> Introduction Example)
-    and combine: "BEM 100: Introduction Example"
+    document title from the PDF meta data and combine, e.g.,:
+    "BEM 100: Introduction Example"
```
KevinJBoyer, line 31: Truing up the comment with the code

app/src/db/migrations/versions/2024_09_03_add_num_splits_and_split_index.py
yoomlam: Thanks! Please add DB comments for the `Column`s.

app/src/db/migrations/versions/2024_09_03_add_num_splits_and_split_index.py
```diff
@@ -0,0 +1,70 @@
...
+    op.alter_column(
+        "document",
+        "dataset",
+        existing_type=sa.TEXT(),
+        nullable=False,
+        comment=None,
```
yoomlam, line 27: Is this removing the DB comment? If so, why?

app/src/db/migrations/versions/2024_09_03_add_num_splits_and_split_index.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
+        "document",
+        "region",
+        existing_type=sa.TEXT(),
+        nullable=True,
+        comment="geographical region of the benefit program",
```
yoomlam, line 56: Is this line needed if we're not changing the DB comment.

app/src/db/migrations/versions/2024_09_03_add_num_splits_and_split_index.py
KevinJBoyer: Good catch, it is. It's removing them because the comments weren't attached in the code definition in document.py. I'll add them in as part of this PR.

yoomlam (approved):

PR #70, opened 2024-09-03
test: Use 707.pdf to test ingest_bem_pdfs.py

yoomlam:
## Ticket

https://nava.slack.com/archives/C06DP498D1D/p1725380936499099?thread_ts=1725375586.922079&cid=C06DP498D1D

The cropped 100.pdf used for testing stripped the outline metadata, which is why I [use a new complete file](https://github.com/navapbc/labs-decision-support-tool/blob/903bab7acb82f7d54b2aac696dcedcdbb053bf8f/app/tests/src/util/test_pdf_utils.py#L23) when testing anything that uses the PDF outline.

## Changes

Use 707.pdf to test ingest_bem_pdfs.py

### DIFF
--- a/app/tests/src/test_ingest_bem_pdfs.py
+++ b/app/tests/src/test_ingest_bem_pdfs.py
@@ -11,25 +11,27 @@ from tests.src.test_ingest_policy_pdfs import doc_attribs
 
 @pytest.fixture
 def policy_s3_file(mock_s3_bucket_resource):
-    data = smart_open("/app/tests/docs/100.pdf", "rb")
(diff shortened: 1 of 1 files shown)


ccheng26 (approved):
no issues with doing this, but should we move the pdf the /docs folder as opposed to having it in utils?

yoomlam:
> no issues with doing this, but should we move the pdf the /docs folder as opposed to having it in utils?

I tried that in a prior PR -- see https://github.com/navapbc/labs-decision-support-tool/pull/57#discussion_r1729267139

PR #71, opened 2024-09-03
fix: Add heuristics and fixes to improve BEM pdf parsing

yoomlam:
## Ticket

- [Heuristic 1](https://nava.slack.com/archives/C06DP498D1D/p1725396917417349?thread_ts=1725395128.529479&cid=C06DP498D1D): if a `ListItem` or `List` ends with a bullet, merge with the next element.

## Changes

- [add more assertions to check text and headings](https://github.com/navapbc/labs-decision-support-tool/commit/8935ba12952994324a311c78cfdf862e227295e9)
- heuristic 1: [merge text ending with bullet with next element](https://github.com/navapbc/labs-decision-support-tool/pull/71/commits/121181c792e9d54cd7fa631ede3d05cf9b32c8f0)
- fix: [unstructured will strip() texts, so we need to strip the styling text as well](https://github.com/navapbc/labs-decision-support-tool/pull/71/commits/7a4897f96d8da29332b07c6dc25c1fc09c334771)
- [fix processing Title elements to not cause warning](https://github.com/navapbc/labs-decision-support-tool/pull/71/commits/3f58d508398f52ad517fb2574a3cd0101e667b51)
- [if a Title element cannot be found as a heading in the outline, include it as a Title element](https://github.com/navapbc/labs-decision-support-tool/pull/71/commits/6dfd8148c656c2bc61ad693fe5d23ad8e344b03a)
   - heuristic 2: [group each Title element with the next element](https://github.com/navapbc/labs-decision-support-tool/pull/71/commits/52591df14751736a60a085064e9b31115588bce6)
- ... and more -- see commit history below.

## Testing

Updated `test__ingest_bem_pdfs`

### DIFF
--- a/app/src/ingest_bem_pdfs.py
+++ b/app/src/ingest_bem_pdfs.py
@@ -100,6 +100,9 @@ def _enrich_texts(file: BinaryIO) -> list[EnrichedText]:
 
     outline: list[Heading] = pdf_utils.extract_outline(file)
     current_headings: list[Heading] = []
+
+    prev_element_was_empty_list_item = False
+
     for element in unstuctured_elem_list:
         if element.category == "Footer" or element.category == "Header":
             continue
@@ -115,6 +118,19 @@ def _enrich_texts(file: BinaryIO) -> list[EnrichedText]:
             if next_heading := _next_heading(outline, element, current_headings):
                 current_headings = next_heading
                 continue
+
+        # Sometimes Unstructured splits a ListItem into an empty ListItem
+        # and then either a NarrativeText,  UncategorizedText, or Title --
+        # for example, BEM 100 page 8 or page 13
+        if element.category == "ListItem" and not element.text:
+            prev_element_was_empty_list_item = True
+            continue
+        if prev_element_was_empty_list_item:
+            if element.category in ("NarrativeText", "UncategorizedText", "Title"):
+                element.category = "ListItem"
+            else:
+                logger.warning(f"Empty list item not followed by NarrativeText, UncategorizedText, or Title, {element.metadata.page_number}")
+            prev_element_was_empty_list_item = False
 
         try:
             enriched_text_item = EnrichedText(
--- a/app/src/ingest_bem_pdfs.py
+++ b/app/src/ingest_bem_pdfs.py
@@ -94,7 +94,7 @@ def _parse_pdf(file: BinaryIO, file_path: str) -> list[EnrichedText]:
 
 
 def _enrich_texts(file: BinaryIO) -> list[EnrichedText]:
-    unstuctured_elem_list = partition_pdf(file=file, strategy="hi_res")
+    unstuctured_elem_list = partition_pdf(file=file, strategy="fast")
     enrich_text_list = []
 
     outline: list[Heading] = pdf_utils.extract_outline(file)
--- a/app/src/ingest_bem_pdfs.py
+++ b/app/src/ingest_bem_pdfs.py
@@ -77,10 +77,14 @@ def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
     enriched_texts = _enrich_texts(file)
     with open("enriched_texts.log", "w") as log_file:
         pprint(enriched_texts, log_file, width=200)
-    stylings = extract_stylings(file)
-    associate_stylings(enriched_texts, stylings)
-    with open("styled_enriched_texts.log", "w") as log_file:
-        pprint(enriched_texts, log_file, width=200)
+    try:
+        stylings = extract_stylings(file)
+        associate_stylings(enriched_texts, stylings)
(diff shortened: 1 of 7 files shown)


app/tests/src/test_ingest_bem_pdfs.py
```diff
@@ -66,15 +66,15 @@ def mock_elements():
     ]
 
 
-@pytest.mark.parametrize("file_location", ["local", "s3"])
+@pytest.mark.parametrize("file_location", ["local"])
```
yoomlam, line 69: Temporarily disable testing `s3`.

app/src/ingest_bem_pdfs.py
```diff
@@ -3,6 +3,7 @@
 import re
 import sys
 import uuid
+from pprint import pprint
```
yoomlam, line 6: Temporarily use `pprint` for debugging

app/src/ingestion/pdf_stylings.py
```diff
@@ -271,7 +271,7 @@ def _create_extracted_text(self, elem: Element) -> ExtractedText | None:
                 if subtype == "/'Footer'":
                     return self._extract_text_in_zone(elem, PageZone.FOOTER)
 
-            logger.info("Ignoring Artifact: %s", elem.toxml())
+            logger.debug("Ignoring Artifact: %s", elem.toxml())
```
yoomlam, line 274: These are often empty artifact tags. 🤷

app/src/ingest_bem_pdfs.py
```diff
@@ -67,13 +68,20 @@ def _ingest_bem_pdfs(
             _add_embeddings(chunks)
             db_session.add_all(chunks)
 
-            _save_json(file_path, chunks)
+            if save_json:
+                # Note that chunks are being added to the DB before saving the JSON.
+                # Originally, we thought about reviewing the JSON manually before adding chunks to the DB.
+                _save_json(file_path, chunks)
```
KevinJBoyer, line 74: I would implement this functionality (being able to manually adjust the json) treating the .pdf.json as caches: by first checking around line ~61 for the presence of a file at`file_path + .json` -- if there is one, load `chunks = json.load(contents_of_saved_json` and go straight to `_add_embeddings(chunks) \ db_session.add_call(chunks)`. So the method would be: ingest the .pdf, then if you don't like how it was loaded, edit the .pdf.json that this produced and re-ingest the pdf (which will then load the stored .pdf.json)

KevinJBoyer: (in a separate PR -- just responding to your comment here)

app/src/ingest_bem_pdfs.py
```diff
@@ -67,13 +68,20 @@ def _ingest_bem_pdfs(
...
 
 
-def _parse_pdf(file: BinaryIO) -> list[EnrichedText]:
+def _parse_pdf(file: BinaryIO, file_path: str) -> list[EnrichedText]:
```
KevinJBoyer, line 77: I'd slightly prefer not to pass in a parameter if it's only being used in a log message -- sequentially this log message follows `logger.info("Processing file: %s", file_path)` to help with tracing (though I agree logging file_path directly in `Failed to extract and associate stylings` is nicer)

app/src/ingestion/pdf_postprocess.py
```diff
@@ -151,6 +163,16 @@ def _group_list_texts(markdown_texts: list[EnrichedText]) -> list[EnrichedText]:
     for current_text in markdown_texts[1:]:
         previous_text = grouped_texts[-1]
 
+        # Unstructured text sometimes splits a bullet from it's text;
```
KevinJBoyer, line 166:
```suggestion
        # Unstructured text sometimes splits a bullet from its text;
```

KevinJBoyer: However I think this is already being done earlier by the code after `# Sometimes Unstructured splits a ListItem into an empty ListItem` above, so we may be able to remove this code

KevinJBoyer (approved):
Great stuff. Left a few comments but nothing blocking, feel free to merge as-is. Thanks for setting up the PR and then getting it ready for merge!

app/src/ingest_bem_pdfs.py
yoomlam: Sounds great!

app/src/ingest_bem_pdfs.py
yoomlam: I concur. Given the exploratory nature of the BEM PDFs, I'll keep this in. For future implementations, passing around a single ParsingContext that includes the filename, current headings, page number, etc. would be helpful.

app/src/ingestion/pdf_postprocess.py
yoomlam: Yea, there's some overlap and it could be cleaned up.

PR #73, opened 2024-09-05
feat: Update citation UI

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-435

## Changes
 - Slightly update LLM prompt:
    - to encourage `(chunk-0)(chunk-1)` style citations (which our code supports) rather than `(chunk-0,1)` (which the LLM invents)
    - to encourage adding citations after punctuation.
 - Remove similarity scores from UI
 - Set defaults for BEM engine similarity score cut off to -1, and match citations shown k with retrieval k (8)
 - Update citations UX to match the Figma prototype (see ticket for screenshot)
 - Add hotfix for proper markdown list rendering by Chainlit

## Context for reviewers
n/a

## Testing
image

### DIFF
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -152,7 +152,7 @@ def _format_to_accordion_group_html(documents: OrderedDict[Document, list[ChunkW
                 </div>
             </div>"""  # noqa: B907
 
-    return "<h3>Source(s)</h3>" + html if html else ""
+    return "\n<h3>Source(s)</h3>" + html if html else ""
 
 
 def _get_bem_url(text: str) -> str:
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -114,7 +114,7 @@ def _format_to_accordion_group_html(documents: OrderedDict[Document, list[ChunkW
             formatted_chunk = _replace_bem_with_link(chunk.content)
 
             # Adjust markdown for lists so Chainlit renders correctly
-            formatted_chunk = formatted_chunk.replace(r"^    - ", "- ")
+            formatted_chunk = re.sub('^    - ', "- ", formatted_chunk, flags=re.MULTILINE)
             if formatted_chunk.startswith("- "):
                 formatted_chunk = "\n" + formatted_chunk
 
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -118,7 +118,9 @@ def _format_to_accordion_group_html(documents: OrderedDict[Document, list[ChunkW
             if formatted_chunk.startswith("- "):
                 formatted_chunk = "\n" + formatted_chunk
 
-            bem_url_for_page = _get_bem_url(document.name) + "#page=" + str(chunk.page_number)
+            bem_url_for_page = _get_bem_url(document.name)
+            if chunk.page_number:
+                bem_url_for_page += "#page=" + str(chunk.page_number)
 
             citation_heading = f"<h4>Citation {citation_number}:</h4>"
             chunk_headings = "<p>" + " → ".join(chunk.headings) + "</p>" if chunk.headings else ""
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -126,7 +126,11 @@ def _format_to_accordion_group_html(documents: OrderedDict[Document, list[ChunkW
             chunk_headings = "<p>" + " → ".join(chunk.headings) + "</p>" if chunk.headings else ""
             citation_body = f'<div class="margin-left-2 border-left-1 border-base-lighter padding-left-2">{formatted_chunk}</div>'
             citation_link = (
-                f"<p><a href={bem_url_for_page!r}>Open document to page {chunk.page_number}</a></p>"
+                (
+                    f"<p><a href={bem_url_for_page!r}>Open document to page {chunk.page_number}</a></p>"
+                )
+                if chunk.page_number
+                else ""
             )
             citations += citation_heading + chunk_headings + citation_body + citation_link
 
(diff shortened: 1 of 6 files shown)


app/src/format.py
```diff
@@ -102,19 +102,41 @@ def _format_to_accordion_html(document: Document, score: float) -> str:
...
+        citation_number_start = citation_number
+
+        for chunk_with_score in documents[document]:
+            chunk = chunk_with_score.chunk
+            formatted_chunk = _replace_bem_with_link(chunk.content)
+
+            # Adjust markdown for lists so Chainlit renders correctly
+            formatted_chunk = formatted_chunk.replace("    - ", "- ")
+            if formatted_chunk.startswith("- "):
+                formatted_chunk = "\n" + formatted_chunk
```
yoomlam, line 119: Note to self: Currently, this can happen because the previous sentence doesn't end with a colon. It can also happen because the list is the first element after a new heading.

app/src/format.py
```diff
@@ -102,19 +102,41 @@ def _format_to_accordion_html(document: Document, score: float) -> str:
...
+
+            bem_url_for_page = _get_bem_url(document.name) + "#page=" + str(chunk.page_number)
```
yoomlam, line 121: In `citations.py` we check `if chunk.page_number else ""`, but in this code block we don't. Is this intentional?

KevinJBoyer: Not intentional, will update this code to do the same!

app/src/format.py
```diff
@@ -102,19 +102,43 @@ def _format_to_accordion_html(document: Document, score: float) -> str:
...
+
+            bem_url_for_page = _get_bem_url(document.name)
+            if chunk.page_number:
+                bem_url_for_page += "#page=" + str(chunk.page_number)
+
+            citation_heading = f"<h4>Citation {citation_number}:</h4>"
+            chunk_headings = "<p>" + " → ".join(chunk.headings) + "</p>" if chunk.headings else ""
+            citation_body = f'<div class="margin-left-2 border-left-1 border-base-lighter padding-left-2">{formatted_chunk}</div>'
+            citation_link = (
+                f"<p><a href={bem_url_for_page!r}>Open document to page {chunk.page_number}</a></p>"
```
yoomlam, line 129: Another use of `chunk.page_number`

KevinJBoyer: Fixed!

app/tests/src/test_citations.py
```diff
@@ -17,9 +17,14 @@ def test_get_context_for_prompt():
 
 
 def test_add_citations():
-    assert add_citations("This is a citation (chunk-0)", []) == "This is a citation (chunk-0)</br>"
+    assert add_citations("This is a citation (chunk-0)", []) == "This is a citation (chunk-0)"
 
     chunks = ChunkFactory.build_batch(2)
+    assert (
+        add_citations("This is a citation (chunk-0) and another (chunk-1).", chunks)
+        == "This is a citation <sup><a href='#'>1</a>&nbsp;</sup> and another <sup><a href='#'>2</a>&nbsp;</sup>."
```
yoomlam, line 25: Trying to understand `<a href='#'>`: should there be an anchor name after `#`?
Found somewhat related https://stackoverflow.com/questions/37640583/understanding-a-href

app/src/format.py
```diff
@@ -102,19 +102,47 @@ def _format_to_accordion_html(document: Document, score: float) -> str:
...
-                f"<p>{_replace_bem_with_link(formatted_chunk)} </p>" if formatted_chunk else ""
+
+        citation_number_start = citation_number
+
+        for chunk_with_score in documents[document]:
+            chunk = chunk_with_score.chunk
+            formatted_chunk = _replace_bem_with_link(chunk.content)
+
+            # Adjust markdown for lists so Chainlit renders correctly
+            formatted_chunk = formatted_chunk.replace("    - ", "- ")
```
yoomlam, line 117: Should this be replacing only the first-level of list items so that nested list items aren't unnested?
```suggestion
            formatted_chunk = formatted_chunk.replace(r"^    - ", "- ")
```

app/tests/src/test_citations.py
KevinJBoyer: It's a placeholder from this line: https://github.com/navapbc/labs-decision-support-tool/blob/kb/improve-citation-ux/app/src/citations.py#L42

We could also update the code to simply not wrap non-BEM citations in links if you prefer!

app/src/format.py
KevinJBoyer: Good call. Ultimately we should fix this in the ingestion code and remove the code here (hopefully before or at the same time as adding support for nested links), but better safe than sorry

app/src/format.py
```diff
@@ -102,19 +102,47 @@ def _format_to_accordion_html(document: Document, score: float) -> str:
...
-                f"<p>{_replace_bem_with_link(formatted_chunk)} </p>" if formatted_chunk else ""
+
+        citation_number_start = citation_number
+
+        for chunk_with_score in documents[document]:
+            chunk = chunk_with_score.chunk
+            formatted_chunk = _replace_bem_with_link(chunk.content)
+
+            # Adjust markdown for lists so Chainlit renders correctly
+            formatted_chunk = formatted_chunk.replace(r"^    - ", "- ")
```
yoomlam, line 117: Sorry, I don't think my earlier suggestion actually works. How about this?
```suggestion
            formatted_chunk = re.sub('^    - ', "- ", formatted_chunk, flags=re.MULTILINE)
```

app/tests/src/test_citations.py
yoomlam: I'm not following, but testing it locally, it works!

yoomlam: Oh, wait. It's b/c `chunk.document.name` is not present.

yoomlam (approved):

ccheng26 (approved):
code changes lgtm

PR #74, opened 2024-09-05
feat: add dash formatting to list items

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-424

## Changes

> separated dash addition to list to its own function

## Context for reviewers

> Noted in 100 and 707, unstructured preserves the  • character in some lists; for sublists, the outer bullet is cut off, but the bullets are preserved in the sublist. The code assumes that a list item, followed by a list item with a bullet will produce a sublist


app/src/ingestion/pdf_postprocess.py
```diff
@@ -114,8 +114,27 @@ def _add_link_markdown(e_text: EnrichedText) -> EnrichedText:
...
+        and prev_e_text.type == TextType.LIST_ITEM
+        and current_e_text.type == TextType.LIST_ITEM
+    ):
+        # if sublist, indent 2 spaces
+        if "\u2022" in current_e_text.text:
+            current_e_text.text = "  - " + current_e_text.text
+        else:
+            current_e_text.text = "- " + current_e_text.text
+    elif current_e_text.type == TextType.LIST_ITEM:
+        current_e_text.text = "- " + current_e_text.text
```
yoomlam, line 131: I think you need to update https://github.com/navapbc/labs-decision-support-tool/blob/771f61395ec8d69fd313e6ce851ce4d57995452b/app/src/ingestion/pdf_postprocess.py#L169 since it's expecting `    - `

yoomlam:
> The code assumes that a list item, followed by a list item with a bullet will produce a sublist

Does Unstructured capture some nested list items as nested under a list item? cc: @KevinJBoyer 
If so, this PR doesn't handle that case.

app/src/ingestion/pdf_postprocess.py
```diff
@@ -114,8 +114,27 @@ def _add_link_markdown(e_text: EnrichedText) -> EnrichedText:
...
+def _add_list_markdown(
+    prev_e_text: EnrichedText | None, current_e_text: EnrichedText
+) -> EnrichedText:
+    if (
+        prev_e_text
+        and prev_e_text.type == TextType.LIST_ITEM
+        and current_e_text.type == TextType.LIST_ITEM
+    ):
+        # if sublist, indent 2 spaces
+        if "\u2022" in current_e_text.text:
```
yoomlam, line 126: Should this be checking for the bullet at the start of the text and not anywhere in the text?

app/tests/src/ingestion/test_pdf_postprocess.py
```diff
@@ -341,3 +354,36 @@ def test__add_link_markdown(caplog):
...
+            page_number=2,
+            stylings=[
+                Styling(
+                    text="• Sub nested item.",
+                    pageno=2,
+                    headings=[Heading(title="Section 1", level=1)],
+                    wider_text="Sub nested item.",
+                    bold=False,
+                )
+            ],
```
yoomlam, line 374: Is `stylings` needed for this test?

app/src/ingestion/pdf_postprocess.py
ccheng26: there's instances where the bullet appears as the second item due to the paragraph blocks

yoomlam: > the second item

Do you mean the second character? If so, use `.strip()`.

yoomlam: Add a unit test that shows "where the bullet appears as the second item".

app/tests/src/ingestion/test_pdf_postprocess.py
```diff
@@ -41,28 +45,37 @@ def test_add_markdown():
         ),
     ]
 
-    result = add_markdown(enriched_texts)
 
-    assert result == [
+@pytest.fixture
```
yoomlam, line 49: What's the reasoning for turning an expected result into a test fixture? https://blog.machinet.net/post/what-is-a-test-fixture-a-comprehensive-guide

app/tests/src/ingestion/test_pdf_postprocess.py
```diff
@@ -341,3 +354,36 @@ def test__add_link_markdown(caplog):
...
+            ],
+        )
+    )
+
+    prev_enriched_text_val = None
+    markdown_text = []
+    for enriched_text in enriched_texts:
+        if prev_enriched_text_val is not None:
+            markdown_text.append(_add_list_markdown(prev_enriched_text_val, enriched_text))
+        prev_enriched_text_val = enriched_text
```
yoomlam, line 383: Curious to see this much logic in a unit test. I understand you're trying to replicate the logic in `add_markdown()` so maybe just call `add_markdown()` with the intent of testing `_add_list_markdown()`? Alternatively, call `_add_list_markdown()` with different inputs:
```
_add_list_markdown(None, enriched_text)
assert ...

_add_list_markdown(prev_enriched_text, enriched_text)
assert ...

etc.
```

yoomlam (changes requested):

app/src/ingestion/pdf_postprocess.py
ccheng26: by second item I mean something like ` Parent/provider certifications. • Day/date. • Children's names. In/out times. `, Parent/provider certifications is part of the list. I wasn't sure if we should do a find/replace on the  •  values since they'll still be inconsistent (In/out times is another bullet on the same list level)

yoomlam: I see. If you're searching for those second bullet items, it makes sense to address them. The current code adds to only the first bullet item: `"  - " + current_e_text.text`. Unit tests would demonstrate the scenarios that you're addressing.

KevinJBoyer (approved):
LGTM with some suggestions around testing clarity. Also if I'm following the changes here right, I don't think we need these two lines anymore: https://github.com/navapbc/labs-decision-support-tool/blob/main/app/src/format.py#L116 is that right? If so, could you remove them as part of this PR?

KevinJBoyer:
@yoomlam whoops sorry I missed your comment when I was reviewing earlier. I'm not sure, we should check. Do you or @ccheng26 have a sublist handy that we could run through the exploration notebook?

yoomlam:
@KevinJBoyer I don't have any nested lists in particular. Here are some:
- 100.pdf, bottom of page 12; I scanned through [the sample](https://github.com/navapbc/labs-decision-support-tool/blob/main/app/notebooks/UnstructoredExploration.ipynb) but didn't see the list nested.
```
[ListItem] A request is called in through administrative chan
[ListItem] A request is called in through administrative chan
[ListItem] • • • • • • A detailed description of the issue. •
[ListItem] A request is sent to the appropriate program polic
```
- 707.pdf, top of page 2

app/tests/src/ingestion/test_pdf_postprocess.py
```diff
@@ -1,8 +1,11 @@
 import logging
 
+import pytest
+
 from src.ingestion.pdf_elements import EnrichedText, Heading, Link, TextType
 from src.ingestion.pdf_postprocess import (
     _add_link_markdown,
+    _add_list_markdown,
```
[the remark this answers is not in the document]
ccheng26: Looks similar but two different functions.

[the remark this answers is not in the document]
KevinJBoyer: 😂 🤦

app/src/format.py
```diff
@@ -113,11 +113,6 @@ def _format_to_accordion_group_html(documents: OrderedDict[Document, list[ChunkW
             chunk = chunk_with_score.chunk
             formatted_chunk = _replace_bem_with_link(chunk.content)
 
-            # Adjust markdown for lists so Chainlit renders correctly
-            formatted_chunk = re.sub("^    - ", "- ", formatted_chunk, flags=re.MULTILINE)
-            if formatted_chunk.startswith("- "):
-                formatted_chunk = "\n" + formatted_chunk
```
yoomlam, line 119: See https://github.com/navapbc/labs-decision-support-tool/pull/73/files#r1745798163. Are you certain this can be removed?

app/tests/src/ingestion/test_pdf_postprocess.py
```diff
@@ -144,32 +152,47 @@ def test_concatenate_list_items():
...
             page_number=1,
+            id=None,
+            stylings=None,
+            links=None,
         ),
         EnrichedText(
             text="New list item in new section with continuing sentence on next page",
             type=TextType.LIST_ITEM,
-            headings=[Heading(title="Section 1", level=1)],
+            headings=[Heading(title="Section 1", level=1, pageno=None)],
```
yoomlam, line 191: Headings should all have page numbers
```suggestion
            headings=[Heading(title="Section 1", level=1, pageno=1)],
```

app/tests/src/ingestion/test_pdf_postprocess.py
```diff
@@ -341,3 +364,42 @@ def test__add_link_markdown(caplog):
...
+            text="• Sub nested item.",
+            type=TextType.LIST_ITEM,
+            headings=[Heading(title="Section 1", level=1)],
+            page_number=2,
+            stylings=None,
+        ),
+    )
+
+    assert first_list_level.text == "- First item."
+    assert second_list_level.text == "  - \n  -  Sub nested item."
```
yoomlam, line 385: This desired result looks odd with a blank list item. Should this be:
```suggestion
    assert second_list_level.text == "    - Sub nested item."
```

This result is probably a result of `current_e_text.text.replace("\u2022", "\n  - ")`. We can pair on this to talk through what we want the result to look like and a good algorithm to achieve it.

app/src/ingestion/pdf_postprocess.py
```diff
@@ -114,8 +114,28 @@ def _add_link_markdown(e_text: EnrichedText) -> EnrichedText:
...
 
+def _add_list_markdown(
+    prev_e_text: EnrichedText | None, current_e_text: EnrichedText
+) -> EnrichedText:
+    if (
+        prev_e_text
+        and prev_e_text.type == TextType.LIST_ITEM
+        and current_e_text.type == TextType.LIST_ITEM
+    ):
+        # if sublist, indent 2 spaces
```
yoomlam, line 125: https://commonmark.org/help/tutorial/10-nestedLists.html says:
> To nest one list within another, indent each item in the sublist by four spaces. 
https://nava.slack.com/archives/C06DP498D1D/p1725550690832049?thread_ts=1725547115.307169&cid=C06DP498D1D.
```suggestion
        # if sublist, indent 4 spaces
```

app/src/format.py
yoomlam: You can check by re-ingesting using this PR and looking at the `chunks` table and querying `content LIKE '    -%'` or `content LIKE '-%'`. Here's what my current DB looks like:
image

ccheng26: The output with the updated code looks correct. double clicking on the entry seems to format the results properly for me and the test cases are passing with the lines removed.
Screenshot 2024-09-06 at 12 55 27 PM

yoomlam: Right, so if there are chunks that start with `-`, do they render correctly in the UI? I think this code was intended to correct some list rendering issues, i.e., the UI wouldn't render as a list item without a prepending `\n`.

app/src/ingestion/pdf_postprocess.py
ccheng26: two spaces also works 
Screenshot 2024-09-06 at 1 19 24 PM

yoomlam: 👍  [Interesting thread](https://github.com/DavidAnson/markdownlint/issues/137).

app/tests/src/ingestion/test_pdf_postprocess.py
yoomlam: Reminder to replace all occurrences of `pageno=None` above with `pageno=1`.

yoomlam (approved):
Pending confirmation of the UI rendering of chunks starting with `- ` mentioned in https://github.com/navapbc/labs-decision-support-tool/pull/74/files#r1747450897

app/src/format.py
ccheng26: I was able to replicate the issue, re-added the code with updates

app/src/ingestion/pdf_postprocess.py
```diff
@@ -114,8 +114,33 @@ def _add_link_markdown(e_text: EnrichedText) -> EnrichedText:
...
+        if "\u2022" in current_e_text.text:
+            current_e_text.text = "  - " + current_e_text.text
+            current_e_text.text = current_e_text.text.replace("\u2022", "\n  - ")
+            word_arr = current_e_text.text.split("\n")
+            filtered_dup_bullets = [
+                char for char in word_arr if char.strip() and char.strip() != "-"
+            ]
```
yoomlam, line 132: Is this more correct naming:
```suggestion
            list_items = current_e_text.text.split("\n")
            nonempty_items = [
                item for item in list_items if item.strip() and item.strip() != "-"
            ]
```

yoomlam: This logic seems very specific to the BEM pdfs but if it works (i.e., nested list items are correctly nested), then great! Basically the logic is: if there is a bullet character, then assume it's a nested list item one level down -- we don't handle deeper nesting.
I'd rerun the ingestion and check for nested items in the chunks to ensure they're correctly formatted.
Report any cases of nested list items that aren't identified as tech debt for a future PR.

PR #75, opened 2024-09-09
feat: add ellipses to start/end of text for chunks

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-429

## Changes

> add ellipses to start/middle/end of text for chunks

## Testing
Screenshot 2024-09-09 at 1 48 29 PM

### DIFF
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -184,10 +184,9 @@ def _add_ellipses(chunk: Chunk) -> str:
     chunk_content = chunk.content
     if chunk.num_splits != 0:
         if chunk.split_index == 0:
-            chunk_content = f"{chunk_content}..."
+            return f"{chunk_content} ..."
         elif chunk.split_index == chunk.num_splits:
-            chunk_content = f"...{chunk_content}"
+            return f"... {chunk_content}"
         else:
-            chunk_content = f"...{chunk_content}..."
-        chunk_content = re.sub(r"\.{4,}", "...", chunk_content)
+            return f"... {chunk_content} ..."
     return chunk_content
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -3,7 +3,7 @@ import random
 import re
 from typing import OrderedDict, Sequence
 
-from src.db.models.document import ChunkWithScore, Document
+from src.db.models.document import Chunk, ChunkWithScore, Document
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


app/src/format.py
```diff
@@ -46,6 +46,7 @@ def _get_bem_documents_to_show(
     # Ordered by the highest score of each chunk associated with the document
     documents: OrderedDict[Document, list[ChunkWithScore]] = OrderedDict()
     for chunk_with_score in chunks_with_scores[:chunks_shown_max_num]:
+        chunk_with_score.chunk = _add_ellipses(chunk_with_score.chunk)
```
yoomlam, line 49: I think you should use a new variable to avoid unintentionally updating the DB record.

app/src/format.py
```diff
@@ -176,3 +177,15 @@ def _replace_bem_with_link(text: str) -> str:
...
+def _add_ellipses(chunk: Chunk) -> Chunk:
+    if chunk.num_splits != 0:
+        if chunk.split_index == 0:
+            chunk.content = f"{chunk.content}..."
+        elif chunk.split_index == chunk.num_splits:
+            chunk.content = f"...{chunk.content}"
+        else:
+            chunk.content = f"...{chunk.content}..."
+        chunk.content = re.sub(r"\.{4,}", "...", chunk.content)
+    return chunk
```
yoomlam, line 191: Don't modify `chunk`, also you can reduce indentation and easier to interpret by `return`ing more:
```suggestion
    if chunk.num_splits == 0:
        return chunk

    if chunk.split_index == 0:
        return f"{chunk.content}..."
    
    if chunk.split_index == chunk.num_splits:
        return f"...{chunk.content}"

    return f"...{chunk.content}..."
```

I don't know why the following is needed -- I interpret this line as replace "...." with "...". Is there a test for this?
```   
    chunk.content = re.sub(r"\.{4,}", "...", chunk.content)
```

yoomlam (changes requested):

app/src/format.py
ccheng26: Yes, the first test case has a period at the end of the line. In cases where there are more than 4 .'s, it'll replace it with the ellipses.

yoomlam: I see. How about joining with a space in between: `" ".join(content, "...")`? I like this better for the end of sentences, such as `Some sentence ending with a period. ...` over `Some sentence ending with a period...` since it's obvious the sentence ended.

yoomlam (approved):

PR #76, opened 2024-09-10
feat: add headings to context

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-439

## Changes

> add headings to context

## Testing

> See unit tests


KevinJBoyer (commented):
I think headings is a list of strings, not a string directly, right?

app/src/citations.py
```diff
@@ -15,7 +15,7 @@
 def get_context_for_prompt(chunks: Sequence[ChunkWithScore]) -> str:
     return "\n\n".join(
         [
-            f"Citation: chunk-{index}\nDocument name: {chunk.chunk.document.name}\nContent: {chunk.chunk.content}"
+            f"Citation: chunk-{index}\nDocument name: {chunk.chunk.document.name}\nContent: {chunk.chunk.content}\nHeadings: {chunk.chunk.headings}"
```
KevinJBoyer, line 18: I think it would make more sense to put the headings before the content rather than after. Not sure if it'll have too big a difference on LLM performance, but that's at least what a human would find intuitive.

ccheng26:
> I think headings is a list of strings, not a string directly, right?

Right, my mistake. I've updated the description for the Chunk object. Was there any delimiter we wanted for the headings?

KevinJBoyer:
Thank you! I used a unicode arrow for the user [here](https://github.com/navapbc/labs-decision-support-tool/pull/73/files#diff-7b9325216877d34c24a05920b08a2f7cfa766207d0b2b05b20cf740d5f96f90fR126), but I'm not sure what the best approach is -- we could also use `>`?

ccheng26:
> Thank you! I used a unicode arrow for the user [here](https://github.com/navapbc/labs-decision-support-tool/pull/73/files#diff-7b9325216877d34c24a05920b08a2f7cfa766207d0b2b05b20cf740d5f96f90fR126), but I'm not sure what the best approach is -- we could also use `>`?

I went with '>' since the unicode arrow was causing some test failures.

app/pyproject.toml
```diff
@@ -30,7 +30,7 @@ pdfminer-six = "^20240706"
 unstructured = {extras = ["pdf", "docx", "pptx"], version = "^0.15.8"}
 nltk = "^3.9.1"
 [tool.poetry.group.dev.dependencies]
-black = "^23.9.1"
+black = "^24.8.0"
```
KevinJBoyer, line 33: Just curious what issue the 23.9.1 version was causing?

KevinJBoyer (approved):
LGTM!

app/pyproject.toml
KevinJBoyer: Ah, I see the cannot parse issues. It might be more straightforward to break that up into two lines of code:

```
joined_headings = " > ".join(...)
context = f"some stuff... {joined_headings} ...other stuff"
```

than update black.

PR #79, opened 2024-09-12
feat: Group text with the same heading

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-437

## Changes

> add code to process text chunks with same heading

## Testing

> run `make ingest-bem-pdfs DATASET_ID=bridges-eligibility-manual BENEFIT_PROGRAM=multiprogram BENEFIT_REGION=Michigan FILEPATH=/app/documents/{file_path}`

PR #80, opened 2024-09-19
fix: Address BEM PDF ingestion error cases with bigger chunks

yoomlam:
## Ticket

[Slack](https://nava.slack.com/archives/C06DP498D1D/p1726761922331469)

PR #79 resulted in bigger chunks and more use of `split_paragraph()` and `split_list()`, which revealed some outlier error cases.

## Changes

- Account for higher token counts per chunk when text is split by adding a 20% fudge factor
- Always merge LIST_ITEMs with the prior text, regardless if it doesn't end with `:`
- Account for lists with no introductory sentence
- Correctly split parsed tables, which are not grouped into a single paragraph

## Testing

Ingest all the BEM pdfs and check chunks

### DIFF
--- a/app/src/ingest_bem_pdfs.py
+++ b/app/src/ingest_bem_pdfs.py
@@ -203,16 +203,26 @@ def _split_into_chunks(document: Document, grouped_texts: list[EnrichedText]) ->
         token_count = len(embedding_model.tokenizer.tokenize(paragraph.text))
         if token_count > embedding_model.max_seq_length:
             # Split the text into chunks of approximately equal length by characters,
-            # which doesn't necessarily mean equal number of tokens, but close enough
-            num_of_splits = math.ceil(token_count / embedding_model.max_seq_length)
-            char_limit_per_split = math.ceil(len(paragraph.text) / num_of_splits)
+            # which doesn't necessarily mean equal number of tokens, but close enough.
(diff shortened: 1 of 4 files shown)


app/src/ingest_bem_pdfs.py
```diff
@@ -203,16 +203,28 @@ def _split_into_chunks(document: Document, grouped_texts: list[EnrichedText]) ->
         token_count = len(embedding_model.tokenizer.tokenize(paragraph.text))
         if token_count > embedding_model.max_seq_length:
             # Split the text into chunks of approximately equal length by characters,
-            # which doesn't necessarily mean equal number of tokens, but close enough
-            num_of_splits = math.ceil(token_count / embedding_model.max_seq_length)
+            # which doesn't necessarily mean equal number of tokens, but close enough.
+            # The arbitrary 1.5 tolerance factor tries to account for higher token counts per chunk when text is split.
+            num_of_splits = math.ceil((token_count * 1.5) / embedding_model.max_seq_length)
```
yoomlam, line 208: I had to bump this up to 1.5 from 1.2 for tables in 541.pdf page 5. This is a quick fix for now, but we'll need to be smarter about grouping too much text together.

KevinJBoyer (approved):
Good stuff, thank you!

PR #81, opened 2024-09-20
feat: Sub-chunk citations

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-438

## Changes
 - Move responsibility for adding citations into the formatters
 - Change the citation format for the LLM from `(chunk-#)` to `(citation-#)`
 - Provide formatters access to the original LLM response so they can apply citations
 - Add new formatter, format_bem_subsections, that splits chunks

## Context for reviewers
 - This PR doesn't group sub-chunk citations from the same BEM document into a single accordion because the citations are numbered according to the order the LLM cites them. E.g., if the LLM cites BEM 100, then BEM 101, then BEM 100 (but a different section), these will be citation #s 1, 2, and 3, and displayed in that order.

## Testing
Screenshot 2024-09-20 at 9 11 18 AM


ccheng26 (approved):

app/src/citations.py
```diff
@@ -3,52 +3,111 @@
...
+    "Example (citation-3)(citation-1), another example (citation-1).", then this function will return
+    [context[3], [context[1]]; citations referencing context[3] should be shown to the user as "1" and
+    citations referencing context[1] should be shown as "2".
+
+    """
+    citation_indices = re.findall(CITATION_PATTERN, response)
+
+    # Maintain order and avoid duplicates by using a list and tracking seen indices
+    seen = set()
+    citations = []
```
yoomlam, line 66: Not as efficient but less code: remove `seen` and just check for `context[idx] in citations`.

yoomlam (commented):
partial review

app/src/citations.py
```diff
@@ -3,52 +3,111 @@
...
+
+    return "\n\n".join(context_list)
+
+
+def get_citation_numbers(
+    context: Sequence[ChunkWithSubsection], response: str
+) -> Sequence[ChunkWithSubsection]:
+    """
+    Map (citation-<index>) in `response`, where index is the index in `context`,
+    to a user-friendly citation number.
```
yoomlam, line 54: This comment sounds inconsistent with the return type of the function, which is a `Sequence[ChunkWithSubsection]`. The comment says it maps to a citation _number_.

app/src/citations.py
```diff
@@ -3,52 +3,111 @@
...
+    for index in citation_indices:
+        idx = int(index)  # Convert string to integer index
+        if 0 <= idx < len(context) and idx not in seen:
+            citations.append(context[idx])
+            seen.add(idx)
+
+    return citations
 
 
 def add_citations(response: str, chunks: list[Chunk]) -> str:
```
yoomlam, line 77: Had to search usage to understand what this was doing.
```suggestion
def reify_citations(response: str, chunks: list[Chunk]) -> str:
    "Given raw LLM response, replace citations with references to corresponding ChunkWithSubsection citations"
```

app/src/citations.py
```diff
@@ -3,52 +3,111 @@
...
-        if index >= len(chunks):
-            logger.warning("LLM generated a citation for a chunk ({index}) that doesn't exist.")
-            return f"(chunk-{index})"
+        if index >= len(context):
+            logger.warning("LLM generated a citation for a reference ({index}) that doesn't exist.")
+            return f"(citation-{index})"
+
+        # Find the matching ChunkWithSubsection in the user-friendly
+        # citation numbers and add 1 to get the citation number to display
+        citation_number = 1 + citation_numbers.index(context[index])
```
yoomlam, line 98: Note to self: this manipulation of the index for user-friendliness gets confusing. In a future refactoring PR, try to separate backend logic from UI user-friendliness.

app/src/format.py
```diff
@@ -12,15 +14,17 @@
...
 
 def format_guru_cards(
     chunks_shown_max_num: int,
     chunks_shown_min_score: float,
     chunks_with_scores: Sequence[ChunkWithScore],
+    raw_response: str,
 ) -> str:
+
+    chunks = [c.chunk for c in chunks_with_scores]
+    response_with_citations = add_citations(raw_response, chunks)
```
yoomlam, line 26: Consider extracting this into a function since it's used multiple times.

yoomlam (changes requested):
Done review

app/src/citations.py
yoomlam: A dictionary result is easier for me to understand.

yoomlam: And fortunately, Python dictionaries retain their insertion order.

app/src/citations.py
```diff
@@ -3,52 +3,111 @@
...
+    """
+    citation_indices = re.findall(CITATION_PATTERN, response)
+
+    # Maintain order and avoid duplicates by using a list and tracking seen indices
+    seen = set()
+    citations = []
+
+    for index in citation_indices:
+        idx = int(index)  # Convert string to integer index
+        if 0 <= idx < len(context) and idx not in seen:
```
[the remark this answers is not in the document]
yoomlam: Please add a code comment to that effect as a reminder for me that the `citation_indices` are coming from the LLM and are susceptible to hallucinations.

app/src/citations.py
KevinJBoyer: Will refactor to use dictionaries shortly!

KevinJBoyer: Done, think that's much easier to follow now. I ended up flipping the keys/values, so the dictionaries now look like:

```
ChunkWithSubsection -> 1
ChunkWithSubsection -> 2
```

etc. -- since most lookups are asking for the right number to display to the user given a particular ChunkWithSubsection.

app/src/citations.py
```diff
@@ -3,52 +3,119 @@
...
         # Leave a citation for chunks that don't exist alone
-        if index >= len(chunks):
-            logger.warning("LLM generated a citation for a chunk ({index}) that doesn't exist.")
-            return f"(chunk-{index})"
+        if index >= len(context):
+            logger.warning("LLM generated a citation for a reference ({index}) that doesn't exist.")
+            return f"(citation-{index})"
+
+        # Find the matching ChunkWithSubsection in the user-friendly
+        # citation numbers and add 1 to get the citation number to display
```
yoomlam, line 98:
```suggestion
        # Get the user-friendly citation number matching ChunkWithSubsection
```

app/src/db/models/document.py
```diff
@@ -76,5 +76,11 @@ class ChunkWithScore:
...
     document: Document
-    # The maxmium similarity score of all Chunks associated with that document
+    # The maximium similarity score of all Chunks associated with that document
     max_score: float
+
+
+@dataclass(frozen=True)
+class ChunkWithSubsection:
+    chunk: Chunk
+    subsection: str
```
yoomlam, line 86:
```suggestion
    # specific substring within chunk.text
    subsection: str
```

app/src/format.py
```diff
@@ -190,3 +235,9 @@ def _add_ellipses(chunk: Chunk) -> str:
         else:
             return f"... {chunk_content} ..."
     return chunk_content
+
+
+def _get_chunks_from_chunks_with_scores(
+    chunks_with_scores: Sequence[ChunkWithScore],
+) -> list[Chunk]:
+    return [c.chunk for c in chunks_with_scores]
```
yoomlam, line 243: Not used?

yoomlam (approved):

app/src/format.py
KevinJBoyer, line 243:
```suggestion
```

PR #83, opened 2024-10-01
feat: Add initial chatbot API

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-473
"Technical investigation for exposing an API"

## Changes

This PR demonstrates how the API _could work_. Since we don't have requirements concerning how the client will interact with the chatbot API, some guesses were made and the easiest path was chosen. I referenced https://navalabs.atlassian.net/browse/DST-461 for initial thoughts.

Added API endpoints using FastAPI:
image

Logs to LiteralAI.

## Testing

Set `LITERAL_API_KEY` env variable.
Start API: `python -m src.chat_api`
It will use port 8001 so as to not conflict with the Chainlit app.

Browse to http://0.0.0.0:8001/docs or run the following:
```
# Check that the API is up
curl -X GET 'http://0.0.0.0:8001/healthcheck'

# Get an API key
CHAT_API_KEY=`curl -X 'GET' 'http://0.0.0.0:8001/api_key/SBNdoe'`; echo $CHAT_API_KEY

# List allowed engines for user
curl -X GET 'http://0.0.0.0:8001/engines?api_key='`echo $CHAT_API_KEY|tr -d '"'`'&username=john'

# Query chatbot, expecting JSON return value
curl -X POST 'http://0.0.0.0:8001/query' -H 'Content-Type: application/json' -d '{ "api_key": '$CHAT_API_KEY', 
"username": "sue", 
"message": "refugee programs?" }'

# Query chatbot, expecting HTML return string
curl -X POST 'http://0.0.0.0:8001/query_html' -H 'Content-Type: application/json' -d '{ "api_key": '$CHAT_API_KEY',
"username": "sue",
"message": "refugee programs?" }'
```

PR #84, opened 2024-10-01
fix ci: Set test coverage thresholds

yoomlam:
## Ticket

[Slack](https://nava.slack.com/archives/C06DP498D1D/p1727795570823389)

## Changes

Set test coverage thresholds.
The PR check will fail if thresholds aren't met.

## Testing

Tested here: https://github.com/navapbc/labs-decision-support-tool/pull/83#issuecomment-2386124218
which now shows a red status.

### DIFF
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -56,3 +56,6 @@ jobs:
           sourceDir: app
           coverageFile: app/coverage_report.xml
           token: ${{ secrets.GITHUB_TOKEN }}
+          thresholdAll: 0.8
+          thresholdNew: 0.9
+          thresholdModified: 0.8


.github/workflows/ci-app.yml
```diff
@@ -56,3 +56,6 @@ jobs:
           sourceDir: app
           coverageFile: app/coverage_report.xml
           token: ${{ secrets.GITHUB_TOKEN }}
+          thresholdAll: 0.8
+          thresholdNew: 0.9
+          thresholdModified: 0.8
```
KevinJBoyer, line 61:
```suggestion
          thresholdModified: 0.8
```

Is this file specific -- in other words, does this mean edits to a file with no pre-existing coverage will fail this check if the PR doesn't take coverage from 0 -> 80%?

KevinJBoyer (approved):
LGTM with one thought that we might want to relax thresholdModified to 0 if I understand how it works correctly

.github/workflows/ci-app.yml
yoomlam: See https://github.com/navapbc/labs-decision-support-tool/pull/83#issuecomment-2386124218

yoomlam: I added an exclusion for app.py:
https://github.com/navapbc/labs-decision-support-tool/pull/83/commits/2f13d8e62102632e5fb8572f4d6ec9aaf2fc75ea

ccheng26 (approved):

.github/workflows/ci-app.yml
yoomlam: So, I'd recommend that files with no pre-existing coverage be added to the `omit` list rather than changing the threshold.

PR #90, opened 2024-10-07
fix: Format markdown to HTML to avoid Chainlit reformatting it

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-507

## Changes

Format markdown to HTML using `markdown` ourselves to avoid Chainlit from reformatting citations and escaping the accordion UI panel

## Testing

See https://navalabs.atlassian.net/browse/DST-507?focusedCommentId=10231

### DIFF
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -70,7 +70,7 @@ def _get_bem_documents_to_show(
 
 def to_html(text: str) -> str:
     # markdown expects '\n' before the start of a list
-    corrected_text = re.sub(r'^- ', "\n- ", text, flags=re.MULTILINE, count=1)
+    corrected_text = re.sub(r"^- ", "\n- ", text, flags=re.MULTILINE, count=1)
     return markdown.markdown(corrected_text)
 
 


app/src/format.py
```diff
@@ -66,6 +66,13 @@ def _get_bem_documents_to_show(
     return documents
 
 
+import markdown
+
+
+def to_html(text: str) -> str:
+    return markdown.markdown(text.replace("- ", "\n- ", 1))
```
yoomlam, line 73: `markdown` expects `\n` before the start of a list.

ccheng26 (approved):
lgtm

app/src/format.py
KevinJBoyer: IIUC this is for situations where the text literally starts with a list -- if that's right, we should check `text.startswith("- ")`, since `- ` is a plausible part of prose (as someone who uses dashes a lot, for example, in this sentence.)

app/src/format.py
```diff
@@ -225,6 +237,7 @@ def _format_to_accordion_group_html(documents: OrderedDict[Document, list[ChunkW
                 {citations}
                 </div>
             </div>"""  # noqa: B907
+    print(html)
```
KevinJBoyer, line 240: Stray `print`

KevinJBoyer (approved):
LGTM with one question around `to_html`

app/src/format.py
```diff
@@ -66,6 +68,12 @@ def _get_bem_documents_to_show(
     return documents
 
 
+def to_html(text: str) -> str:
+    # markdown expects '\n' before the start of a list
+    corrected_text = re.sub(r'^- ', "\n- ", text, flags=re.MULTILINE, count=1)
+    return markdown.markdown(corrected_text)
```
yoomlam, line 74: @KevinJBoyer How about this?

PR #92, opened 2024-10-09
refactor: Citations, subsections, chunks

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-454

## Changes

Refactor the current code to align more closely with separation of chunks and citations (https://navasage.atlassian.net/wiki/spaces/NL/pages/1342177290/Tech+Spec+Separating+Chunks+and+Citations) so that the code is more customizable to different types of data sources.

## Testing

No behavior changes.


app/src/generate.py
```diff
@@ -51,34 +48,27 @@ def get_models() -> dict[str, str]:
...
+    ]
 
-    if context:
-        context_text = create_prompt_context(context)
-        messages = [
-            {
-                "content": PROMPT,
-                "role": "system",
-            },
+    if context_text:
```
yoomlam, line 63: Minor refactoring to differentiate the returned value when `context_text` is present.

app/src/citations.py
```diff
@@ -43,78 +61,81 @@ def create_prompt_context(chunks: Sequence[Chunk]) -> str:
...
-    return citations
+# TODO: move this to format.py
+def reify_citations(response: str, subsections: Sequence[ChunkWithSubsection]) -> str:
+    remapped_citations = remap_citation_ids(subsections, response)
+    return add_citation_links(response, remapped_citations)
 
 
-def reify_citations(response: str, chunks: list[Chunk]) -> str:
+# TODO: move this to format.py since this is UI logic
+def add_citation_links(response: str, remapped_citations: Sequence[ChunkWithSubsection]) -> str:
```
yoomlam, line 109: Thoughts about moving `reify_citations()` and `add_citation_links()` to `format.py` since this is UI logic?

app/src/generate.py
```diff
@@ -1,23 +1,20 @@
...
 Keep your answers as similar to your knowledge text as you can
 
 When referencing the context, do not quote directly.
-Use the provided citation numbers (e.g., (citation-0)) to indicate when you are drawing from the context.
-To cite multiple sources at once, you can append citations like so: (citation-0)(citation-1), etc.
+Use the provided citation numbers (e.g., (citation-1)) to indicate when you are drawing from the context.
+To cite multiple sources at once, you can append citations like so: (citation-1)(citation-2), etc.
 Place the citations after any closing punctuation for the sentence.
-For example: 'This is a sentence that draws on information from the context.(citation-0)'
+For example: 'This is a sentence that draws on information from the context.(citation-1)'
```
yoomlam, line 17: LLM generated response that cited nonexistent `(citation-0)`. This change fixed that issue.
See https://nava.slack.com/archives/C06DP498D1D/p1728502233060729

yoomlam: If this problem recurs, we could create another `CitationIdFactory` that provides random citation ids to maybe prevent hallucinations.

app/src/citations.py
```diff
@@ -1,38 +1,63 @@
...
-CITATION_PATTERN = r"\(citation-(\d+)\)"
+CITATION_PATTERN = r"\((citation-\d+)\)"
 
 
-def split_into_subsections(
-    chunks: Sequence[Chunk], delimiter: str = "\n\n"
-) -> Sequence[ChunkWithSubsection]:
-    # Given a list of chunks, split them into a flat list of subsections
-    context_mapping = []
+class CitationIdFactory:
```
KevinJBoyer, line 18: What's the purpose of CitationIdFactory -- to give us flexibility to change ID-generation strategies if we want to move away from `1`, `2`, etc. in the future?

KevinJBoyer (approved):
Whoops meant to hit approve

app/src/citations.py
yoomlam: Yes, for flexibility. In case we want to start from 0 or 1 or 100, use a different prefix (other than `citation-`), create random IDs, etc.

RE: naming of `CitationIdFactory`. LOL, it was originally `CitationFactory`. All it's doing is assigning a new ID to the citation. Let me ponder this.

yoomlam: Since this factory is now the only location where `ChunkWithSubsection` is created, it makes sense to call it `CitationFactory`, esp. since a "subsection" is semantically the same as a "citation".

ccheng26 (approved):

PR #93, opened 2024-10-09
feat: Group citations from the same document into one accordion

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-443

## Changes

> Added code to group citation items with same document

## Testing
Enter query `what is household income eligibility?`
Screenshot 2024-10-09 at 11 58 17 AM
Screenshot 2024-10-09 at 11 58 09 AM


app/src/format.py
```diff
@@ -87,42 +92,52 @@ def format_bem_subsections(
...
+                        aria-controls="a-{_accordion_id}">
+                        {",".join(citation_numbers)}. {document.name}
+                    </button>
+                </h4>
+                <div id="a-{_accordion_id}" class="usa-accordion__content usa-prose" hidden>
+                    {citation_headings}
+                    {formatted_citation_body}
+                    {citation_link}
+                </div>
+            </div>"""
```
yoomlam, line 140: I believe variations of this block appears multiple times in this file. Refactor this into a method that is called with arguments to account for the variations.

app/src/format.py
```diff
@@ -87,42 +92,52 @@ def format_bem_subsections(
...
-            <div id="a-{_accordion_id}" class="usa-accordion__content usa-prose" hidden>
-                {citation_headings}
-                {citation_body}
-                {citation_link}
-            </div>
-        </div>"""
+
+    citations_by_document = combine_citations_by_document(citation_to_numbers)
+    for document, chunk_list in citations_by_document.items():
+        for chunk in chunk_list:
```
yoomlam, line 99: Rename `chunk` and `chunk_list` so that they better reflect the types of those variables.

yoomlam:
Testing with `refugee benefits?`, the "Source(s)" accordions are not ordered by citation number:
image

app/src/citations.py
```diff
@@ -75,6 +76,37 @@ def dereference_citations(
     return citations
 
 
+CitationNumberAndSubstring: TypeAlias = dict[int, str]
+ChunkWithCitation: TypeAlias = dict[Chunk, list[CitationNumberAndSubstring]]
```
yoomlam, line 80: FYI: You don't need to `import TypeAlias` but it doesn't hurt:
```suggestion
CitationNumberAndSubstring = dict[int, str]
ChunkWithCitation = dict[Chunk, list[CitationNumberAndSubstring]]
```

app/src/citations.py
```diff
@@ -75,6 +76,37 @@ def dereference_citations(
...
+    citations: dict[ChunkWithSubsection, int]
+) -> dict[Document, list[ChunkWithCitation]]:
+    """
+    This function groups the chunks by document and nests the values of the citation number and
+    substring.
+
+    Args:
+        citations (dict[ChunkWithSubsection, int])
+        ChunkWithSubsection is the Chunk item with subsection string
+        and int is the user friendly citation number of that substring chunk
```
yoomlam, line 93: Let's use `subsection` to avoid adding another synonym.
```suggestion
        and int is the user-friendly citation number of that subsection
```

app/src/citations.py
```diff
@@ -75,6 +76,37 @@ def dereference_citations(
...
+    """
+
+    citations_by_chunk: dict[Chunk, list[CitationNumberAndSubstring]] = defaultdict(list)
```
yoomlam, line 96:
```suggestion
    citations_by_chunk: dict[Chunk, list[CitationNumberAndSubsection]] = defaultdict(list)
```

app/src/citations.py
```diff
@@ -75,6 +75,32 @@ def dereference_citations(
...
+    citations_by_document: dict[Document, list[dict[Chunk, list[dict[int, str]]]]] = {}
+    for (
+        chunk,
+        citation_item,
+    ) in citations_by_chunk.items():
+        if chunk.document in citations_by_document:
+            citations_by_document[chunk.document].append({chunk: citation_item})
+        else:
+            citations_by_document[chunk.document] = [{chunk: citation_item}]
+    return citations_by_document
```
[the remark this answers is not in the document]
ccheng26: I opted for readability in this case as opposed to having something like `        citations_by_document[citation.chunk.document].append({citation.chunk: citations_by_chunk[citation.chunk]})`

yoomlam (approved):

PR #102, opened 2024-10-17
refactor: Add ingest_util.tokenize()

yoomlam:
## Ticket

[Slack](https://nava.slack.com/archives/C06DP498D1D/p1729194497470359)

## Changes

Account for start and end tokens when tokenizing and counting tokens.

## Testing

`sentence_transformer.tokenizer.tokenize(split, add_special_tokens=True)` adds the token padding `'<s>'` and `'</s>'`:
```
['<s>', '#', 'unemployment', 'insurance', '–', 'forms', 'and', 'publications', '#', '#', 'publications', '[', 'de', '127', '##5', '##b', ']', '(', 'https', ':', '/', '/', 'ed', '##d', '.', 'ca', '.', 'gov', '/', 'site', '##asse', '##ts', '/', 'files', '/', 'pdf', '_', 'pub', '_', 'ct', '##r', '/', 'de', '##12', '##75', '##b', '.', 'pdf', ')', '–', 'english', '|', '[', 'de', '127', '##5', '##b', '/', 'a', ']', '(', 'https', ':', '/', '/', 'ed', '##d', '.', 'ca', '.', 'gov', '/', 'site', '##asse', '##ts', '/', 'files', '/', 'pdf', '_', 'pub', '_', 'ct', '##r', '/', 'de', '##12', '##75', '##ba', '.', 'pdf', ')', '–', 'armenian', '</s>']
```

### DIFF
--- a/app/src/util/ingest_utils.py
+++ b/app/src/util/ingest_utils.py
@@ -4,6 +4,7 @@ from logging import Logger
 from typing import Callable
 
 from sqlalchemy import delete, select
+from transformers import PreTrainedTokenizerFast
 
 from src.adapters import db
 from src.app_config import app_config
@@ -58,3 +59,18 @@ def process_and_ingest_sys_args(argv: list[str], logger: Logger, ingestion_call:
         db_session.commit()
 
     logger.info("Finished processing")
+
+
+def tokenize(text: str) -> list[str]:
+    """
+    The add_special_tokens argument is specified in PreTrainedTokenizerFast.encode_plus(), parent class of MPNetTokenizerFast.
+    It defaults to True for encode_plus() but defaults to False for .tokenize().
+    Setting add_special_tokens=True will add the special tokens CLS(0) and SEP(2) to the beginning and end of the input text.
+    """
+    tokenizer = app_config.sentence_transformer.tokenizer
+    if isinstance(tokenizer, PreTrainedTokenizerFast):
+        return tokenizer.tokenize(text, add_special_tokens=True)
+    elif tokenizer.__class__.__name__ == "MockTokenizer":
+        return tokenizer.tokenize(text)
+
+    raise ValueError(f"Unexpected tokenizer: {tokenizer.__class__}")
--- a/app/src/ingest_bem_pdfs.py
+++ b/app/src/ingest_bem_pdfs.py
@@ -18,7 +18,7 @@ from src.ingestion.pdf_postprocess import add_markdown, associate_stylings, grou
 from src.ingestion.pdf_stylings import extract_stylings
 from src.util import pdf_utils
 from src.util.file_util import get_files
-from src.util.ingest_utils import process_and_ingest_sys_args
+from src.util.ingest_utils import process_and_ingest_sys_args, tokenize
 from src.util.pdf_utils import Heading
 from src.util.string_utils import split_list, split_paragraph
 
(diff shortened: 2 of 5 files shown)


ccheng26 (approved):

app/src/util/ingest_utils.py
```diff
@@ -58,3 +59,18 @@ def process_and_ingest_sys_args(argv: list[str], logger: Logger, ingestion_call:
...
+    It defaults to True for encode_plus() but defaults to False for .tokenize().
+    Setting add_special_tokens=True will add the special tokens CLS(0) and SEP(2) to the beginning and end of the input text.
+    """
+    tokenizer = app_config.sentence_transformer.tokenizer
+    if isinstance(tokenizer, PreTrainedTokenizerFast):
+        return tokenizer.tokenize(text, add_special_tokens=True)
+    elif tokenizer.__class__.__name__ == "MockTokenizer":
+        return tokenizer.tokenize(text)
+
+    raise ValueError(f"Unexpected tokenizer: {tokenizer.__class__}")
```
KevinJBoyer, line 76: What's the reason for preferring this over calling `sentence_transformer.tokenizer.tokenize(text, add_special_tokens=True) directly?

app/src/util/ingest_utils.py
```diff
@@ -58,3 +59,18 @@ def process_and_ingest_sys_args(argv: list[str], logger: Logger, ingestion_call:
...
+    """
+    The add_special_tokens argument is specified in PreTrainedTokenizerFast.encode_plus(), parent class of MPNetTokenizerFast.
+    It defaults to True for encode_plus() but defaults to False for .tokenize().
+    Setting add_special_tokens=True will add the special tokens CLS(0) and SEP(2) to the beginning and end of the input text.
+    """
+    tokenizer = app_config.sentence_transformer.tokenizer
+    if isinstance(tokenizer, PreTrainedTokenizerFast):
+        return tokenizer.tokenize(text, add_special_tokens=True)
+    elif tokenizer.__class__.__name__ == "MockTokenizer":
+        return tokenizer.tokenize(text)
```
KevinJBoyer, line 74: Suggest instead we modify the mock to accept the `add_special_tokens=True` argument rather than detect it in the prod code

app/src/util/ingest_utils.py
yoomlam: `tokenize()` is called in several places and the `add_special_token` argument is dependent on the specified `sentence_transformer.tokenizer`. Rather than remember to update all `tokenize()` calls, we update in one place depending on the tokenizer.

app/src/util/ingest_utils.py
yoomlam: Does the MockTokenizer have/use special tokens?

KevinJBoyer: We can make it behave that way -- that's better than trying to catch the class instance here.

yoomlam: It's not clear to me how it's better but I'll update MockTokenizer.

yoomlam: > it gives future maintainers the appearance of type safety

I wouldn't interpret that the code is typesafe, but rather that the code only works for those explicit types.

KevinJBoyer: > I wouldn't interpret that the code is typesafe, but rather that the code only works for those explicit types.

Right, but we should try to _make_ it type safe: if someone writes buggy code that passes in a variable not of those types, it's better for that code to fail a linter check than to fail at runtime.

KevinJBoyer: In other words, if the intent of the code is "this only works for these two types," that should be communicated through the type system, which will enforce that intent before code gets merged, than through logical code, which won't fail until runtime.

KevinJBoyer (approved):
Nice work! Thanks for tackling this to make the codebase more robust!

PR #103, opened 2024-10-17
refactor: Move add_embeddings() to ingest_utils.py

yoomlam:
## Ticket

In preparation for https://navalabs.atlassian.net/browse/DST-535

## Changes

Move `add_embeddings()` to `ingest_utils.py` and use new `tokenize()` method.

`add_embeddings()` will be used in `ingest_edd_web.py` in a future PR.
The current `ingest_edd_web.py` does not assert `token_count <= embedding_model.max_seq_length` so `chunk.content` was being truncated in the embedding.

## Testing

No behavior change.

### DIFF
--- a/app/src/util/ingest_utils.py
+++ b/app/src/util/ingest_utils.py
@@ -4,6 +4,7 @@ from logging import Logger
 from typing import Callable
 
 from sqlalchemy import delete, select
+from transformers import PreTrainedTokenizerFast
 
 from src.adapters import db
 from src.app_config import app_config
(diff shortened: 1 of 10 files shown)


KevinJBoyer (approved):
🙏

app/src/util/ingest_utils.py
```diff
@@ -4,10 +4,11 @@
 from typing import Callable
 
 from sqlalchemy import delete, select
+from transformers import PreTrainedTokenizerFast
```
yoomlam, line 7:
```suggestion
```

PR #105, opened 2024-10-18
feat: Add chunk.headings and include in embeddings

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-535

## Changes

- Start a new chunk for each new heading section
    - Add and use `split_markdown_by_heading()` to string_utils.py
- Set chunk.headings so it can be displayed in the UI
- Add and use `headings_as_markdown()` to include the headings as context to include in the embedding
- Ensure no truncation occurs when embedding by `assert`ing token count is under the limit
    - Remove hyperlinks in the text that is used to create embeddings to reduce the token count
- Add headings to UI

## Testing

Added unit tests.

```sh
make init

DEBUG_SCRAPINGS=true poetry run scrape-edd-web

make ingest-edd-web DATASET_ID=ca-edd-web BENEFIT_PROGRAM=employment BENEFIT_REGION=California FILEPATH=src/ingestion/edd_scrapings.json
```

image

### DIFF
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -159,6 +159,7 @@ def format_web_subsections(
     for _, citation in remapped_citations.items():
         _accordion_id += 1
         chunk = citation.chunk
+        citation_headings = " → ".join(chunk.headings) if chunk.headings else ""
         formatted_subsection = to_html(citation.text)
 
         citation_link = ""
@@ -179,6 +180,7 @@ def format_web_subsections(
                 </button>
             </h4>
             <div id="a-{_accordion_id}" class="usa-accordion__content usa-prose" hidden>
+                <div><b>{citation_headings}</b></div>
                 <div class="margin-left-2 border-left-1 border-base-lighter padding-left-2">{formatted_subsection}</div>
                 {citation_link}
             </div>
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -159,7 +159,9 @@ def format_web_subsections(
     for _, citation in remapped_citations.items():
         _accordion_id += 1
         chunk = citation.chunk
-        citation_headings = " → ".join(chunk.headings) if chunk.headings else ""
+        citation_headings = (
+            f"<div><b>{' → '.join(chunk.headings)}</b></div>" if chunk.headings else ""
+        )
         formatted_subsection = to_html(citation.text)
 
         citation_link = ""
@@ -180,7 +182,7 @@ def format_web_subsections(
                 </button>
             </h4>
             <div id="a-{_accordion_id}" class="usa-accordion__content usa-prose" hidden>
-                <div><b>{citation_headings}</b></div>
+                {citation_headings}
                 <div class="margin-left-2 border-left-1 border-base-lighter padding-left-2">{formatted_subsection}</div>
                 {citation_link}
             </div>
--- a/app/src/ingest_edd_web.py
+++ b/app/src/ingest_edd_web.py
@@ -56,7 +56,7 @@ class SplitWithContextText:
             return True
         return False
 
-    def valid(self) -> bool:
+    def is_valid(self) -> bool:
         if self.token_count > app_config.sentence_transformer.max_seq_length:
             logger.warning(
                 "Text too long with %i tokens: %s", self.token_count, self.text_to_encode
@@ -147,7 +147,7 @@ def _split_heading_section(headings: Sequence[str], text: str) -> list[SplitWith
         logger.debug("\n".join([f"[Split {i}]: {s.text_to_encode}" for i, s in enumerate(splits)]))
 
     # Temporary: remove splits that are too long; TODO: fix in a separate PR
-    valid_splits = [split for split in splits if split.valid()]
+    valid_splits = [split for split in splits if split.is_valid()]
     return valid_splits
 
 
--- a/app/src/ingest_edd_web.py
+++ b/app/src/ingest_edd_web.py
@@ -137,7 +137,9 @@ def _split_heading_section(headings: Sequence[str], text: str) -> list[SplitWith
         logger.debug("\n".join([f"[Split {i}]: {s.text_to_encode}" for i, s in enumerate(splits)]))
 
     for split in splits:
-        assert split.valid()
+        assert (
+            split.valid()
+        ), f"token_count: {split.token_count} > {app_config.sentence_transformer.max_seq_length}"
     return splits
 
 
--- a/app/src/ingest_edd_web.py
+++ b/app/src/ingest_edd_web.py
@@ -9,7 +9,7 @@ from src.adapters import db
 from src.app_config import app_config
 from src.db.models.document import Chunk, Document
 from src.util.ingest_utils import add_embeddings, process_and_ingest_sys_args, tokenize
-from src.util.string_utils import headings_as_markdown, remove_links, split_markdown_by_heading
+from src.util.string_utils import remove_links, split_markdown_by_heading
 
 logger = logging.getLogger(__name__)
 
@@ -120,7 +120,7 @@ MarkdownHeaderTextSplitter_DELIMITER = "  \n"
 
 def _split_heading_section(headings: Sequence[str], text: str) -> list[SplitWithContextText]:
     # Add headings to the context_str; other context can also be added
-    context_str = headings_as_markdown(headings)
+    context_str = "\n".join(headings)
     logger.debug("New heading: %s", headings)
 
     splits: list[SplitWithContextText] = []
--- a/app/src/ingest_edd_web.py
+++ b/app/src/ingest_edd_web.py
@@ -7,7 +7,7 @@ from smart_open import open as smart_open
 from src.adapters import db
 from src.app_config import app_config
 from src.db.models.document import Chunk, Document
-from src.util.ingest_utils import process_and_ingest_sys_args
+from src.util.ingest_utils import process_and_ingest_sys_args, tokenize
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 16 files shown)


app/tests/src/test_ingest_edd_web.py
```diff
@@ -75,7 +75,7 @@ def test__ingest_edd(
     caplog, app_config, db_session, edd_web_local_file, edd_web_s3_file, file_location
 ):
     # Force a short max_seq_length to test chunking
-    app_config_for_test.sentence_transformer.max_seq_length = 25
+    app_config_for_test.sentence_transformer.max_seq_length = 50
```
yoomlam, line 78: Needed room for 48 tokens to include nested heading in the embedding.

app/tests/src/util/test_string_utils.py
```diff
@@ -50,6 +57,44 @@ def test_resolve_urls_scenario2():
...
+    heading_sections = list(split_markdown_by_heading(markdown))
+    assert heading_sections == [
+        ((), "Intro text"),
+        (("Heading 1",), "Some text under heading 1"),
+        (("Heading 1", "Heading 2"), "Some text under heading 2"),
+        (("Heading 1", "Heading 2", "Heading 3"), "Some text under heading 3"),
+    ]
+    assert len(heading_sections[0][0]) == 0
+    assert len(heading_sections[1][0]) == 1
+    assert len(heading_sections[2][0]) == 2
```
KevinJBoyer, line 84: I don't think you need these since you're doing an exact comparison on line 76
```suggestion
```

app/src/util/string_utils.py
```diff
@@ -125,3 +126,41 @@ def resolve_urls(base_url: str, markdown: str) -> str:
...
+MARKDOWN_METADATA_KEYS = [key for _, key in MARKDOWN_HEADER_TUPLES]
+
+
+def split_markdown_by_heading(markdown: str) -> Iterator[tuple[Sequence[str], str]]:
+    markdown_splitter = MarkdownHeaderTextSplitter(MARKDOWN_HEADER_TUPLES)
+    for doc in markdown_splitter.split_text(markdown):
+        headings = [
+            # Strip out the markdown link syntax from the headings
+            (
+                re.sub(r"\[([^\]]+)\]\([^\)]+\)", r"\1", doc.metadata[key])
```
KevinJBoyer, line 149: Could you use remove_links here?

app/tests/src/util/test_string_utils.py
```diff
@@ -50,6 +57,44 @@ def test_resolve_urls_scenario2():
...
+    assert headings_as_markdown(["Heading 1"]) == "# Heading 1"
+
+    assert (
+        headings_as_markdown(["Heading 1", "Heading 2", "Heading 3"])
+        == "# Heading 1\n## Heading 2\n### Heading 3"
+    )
+
+
+def test_remove_links():
+    markdown = "[This is a link](https://example.com/relative/path) and [another](https://example.com/absolute/path)"
```
KevinJBoyer, line 99: I think this may have been the intent:
```suggestion
    markdown = "[This is a link](/relative/path) and [another](https://example.com/absolute/path)"
```

app/src/util/string_utils.py
```diff
@@ -125,3 +126,41 @@ def resolve_urls(base_url: str, markdown: str) -> str:
...
+            for key in MARKDOWN_METADATA_KEYS
+        ]
+        # Remove empty headings at the end of list
+        while headings and not headings[-1]:
+            headings.pop()
+        yield tuple(headings), doc.page_content
+
+
+def headings_as_markdown(headings: Sequence[str]) -> str:
+    return "\n".join(f"{"#" * i} {h}" for i, h in enumerate(headings, start=1) if h)
```
KevinJBoyer, line 162: Slick 😎

app/src/ingest_edd_web.py
```diff
@@ -20,55 +22,133 @@ def _ingest_edd_web(
...
+        new_text_to_encode = f"{self.text_to_encode}{delimiter}{remove_links(paragraph)}"
+        token_count = len(tokenize(new_text_to_encode))
+        if token_count <= app_config.sentence_transformer.max_seq_length:
+            self.text += f"{delimiter}{paragraph}"
+            self.text_to_encode = new_text_to_encode
+            self.token_count = token_count
+            return True
+        return False
+
+    def valid(self) -> bool:
```
KevinJBoyer, line 59:
```suggestion
    def is_valid(self) -> bool:
```

KevinJBoyer (approved):
Nice! Left some suggestions/comments, LGTM

app/src/ingest_edd_web.py
```diff
@@ -20,55 +22,133 @@ def _ingest_edd_web(
...
+
+
+# MarkdownHeaderTextSplitter splits text by "\n" then calls aggregate_lines_to_chunks() to reaggregate
+# paragraphs using "  \n", so "\n\n" is replaced by "  \n"
+MarkdownHeaderTextSplitter_DELIMITER = "  \n"
+
+
+def _split_heading_section(headings: Sequence[str], text: str) -> list[SplitWithContextText]:
+    # Add headings to the context_str; other context can also be added
+    context_str = headings_as_markdown(headings)
```
KevinJBoyer, line 123: A stray though, but if `context_str` is only ever intended for generating the embedding, we may not need to render these to markdown (since I don't think the markdown would have any impact on search besides generating additional tokens in the embedding)

app/src/format.py
```diff
@@ -179,6 +180,7 @@ def format_web_subsections(
                 </button>
             </h4>
             <div id="a-{_accordion_id}" class="usa-accordion__content usa-prose" hidden>
+                <div><b>{citation_headings}</b></div>
```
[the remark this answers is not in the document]
yoomlam: Done

app/src/util/string_utils.py
yoomlam: Good point!

app/src/ingest_edd_web.py
```diff
@@ -20,55 +22,133 @@ def _ingest_edd_web(
...
+
+    if len(splits) == 1:
+        logger.info("Heading section fits in 1 chunk: %s", headings[-1])
+    else:
+        logger.info("Split %i has %i tokens", len(splits), splits[-1].token_count)
+        logger.info("Partitioned heading section into %i splits", len(splits))
+        logger.debug("\n".join([f"[Split {i}]: {s.text_to_encode}" for i, s in enumerate(splits)]))
+
+    # Temporary: remove splits that are too long; TODO: fix in a separate PR
+    valid_splits = [split for split in splits if split.valid()]
```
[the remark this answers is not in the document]
yoomlam: If they're too long, then the subsequent call to `add_embeddings()` will fail an `assert chunk.tokens <= embedding_model.max_seq_length`. This temporary code is addressed/removed in #107, ready for review.

app/src/ingest_edd_web.py
yoomlam: I rendered it as markdown because many LLM an interpret markdown. I'm assuming it distinguishes heading text from regular text and puts some more significance on the heading text -- I could be totally wrong, but I played it safe and added the few extra `#`s for headings.

app/src/util/ingest_utils.py
```diff
@@ -71,17 +71,28 @@ def tokenize(text: str) -> list[str]:
...
+) -> None:
     embedding_model = app_config.sentence_transformer
 
+    if texts_to_encode:
+        to_encode = [
+            text if text else chunk.content
+            for chunk, text in zip(chunks, texts_to_encode, strict=True)
+        ]
+    else:
+        to_encode = [chunk.content for chunk in chunks]
```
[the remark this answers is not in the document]
yoomlam: I did intend for `chunks` and `texts_to_encode` to be the same length. I'll add a code comment. I thought about adding an assertion but `zip(..., strict=True)` basically does the same thing.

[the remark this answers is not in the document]
KevinJBoyer: Then why `text if text else chunk.content` -- text can never be None, right?

app/src/ingest_edd_web.py
KevinJBoyer: But at the moment `context_str` is only going to `SplitWithContextText.text_to_encode`, i.e., it'll never be seen by the LLM, right?

app/src/util/ingest_utils.py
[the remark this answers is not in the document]
yoomlam: Does this explain it sufficiently?
https://github.com/navapbc/labs-decision-support-tool/blob/098b0399d5896253f72533cecc33d05a78b640cf/app/src/util/ingest_utils.py#L74-L82

app/src/ingest_edd_web.py
yoomlam: Oh right! I got it mixed up. Want me to replace it with this?
```suggestion
    context_str = "\n".join(headings)
```

KevinJBoyer: That makes sense to me! Might save a few tokens.

app/src/util/ingest_utils.py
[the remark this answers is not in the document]
KevinJBoyer: Yes, thanks for bearing with me!

app/src/ingest_edd_web.py
yoomlam: Replaced.

app/src/util/ingest_utils.py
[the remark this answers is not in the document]
yoomlam: Of course! The code is better for it.

PR #106, opened 2024-10-18
fix: Set CA EDD Web as default engine

KevinJBoyer:
## Ticket

n/a -- item from retro

## Changes
 - Change default from Guru SNAP (which we don't use much anymore) to CA EDD Web

## Context for reviewers
Item from retro

### DIFF
--- a/app/src/app_config.py
+++ b/app/src/app_config.py
@@ -44,6 +44,3 @@ class AppConfig(PydanticBaseEnvConfig):
 
 
 app_config = AppConfig()
-
-app_config.sentence_transformer.tokenizer
-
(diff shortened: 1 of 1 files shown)


app/src/app_config.py
```diff
@@ -44,3 +44,6 @@ def sentence_transformer(self) -> SentenceTransformer:
 
 
 app_config = AppConfig()
+
+app_config.sentence_transformer.tokenizer
+
```
yoomlam, line 49: Stray code?
```suggestion
```

yoomlam (approved):

app/src/app_config.py
KevinJBoyer: Oh yikes, that'll teach me to not read through even "one line" PRs before submitting them!

KevinJBoyer, line 49:
```suggestion
```

PR #110, opened 2024-10-22
fix: Rejoin headings with the subsequent first paragraph

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-510

## Changes

Rejoin headings with the subsequent first paragraph so that headings are not cited by themselves.

## Testing

Updated unit test

### DIFF
--- a/app/src/citations.py
+++ b/app/src/citations.py
@@ -26,7 +26,24 @@ citation_factory = CitationFactory()
 
 
 def default_chunk_splitter(chunk: Chunk) -> list[str]:
-    return chunk.content.split("\n\n")
+    splits = [split for split in chunk.content.split("\n\n") if split]
+    # Rejoin headings with the subsequent first paragraph
+    better_splits = []
+    skip_next = False
+    for i, split in enumerate(splits):
+        if skip_next:
+            skip_next = False
+            continue
+
+        next_split = splits[i + 1] if i + 1 < len(splits) else None
+        if next_split and split.startswith("#") and not next_split.startswith("#"):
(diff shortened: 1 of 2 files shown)


app/src/citations.py
```diff
@@ -26,7 +26,24 @@ def create_citation(self, chunk: Chunk, subsection: str) -> Subsection:
 
 
 def default_chunk_splitter(chunk: Chunk) -> list[str]:
```
yoomlam, line 28: `default_chunk_splitter` is used by `split_into_subsections()` which is used in `BaseEngine`.

app/src/citations.py
```diff
@@ -26,7 +26,24 @@ def create_citation(self, chunk: Chunk, subsection: str) -> Subsection:
...
+            skip_next = False
+            continue
+
+        next_split = splits[i + 1] if i + 1 < len(splits) else None
+        if next_split and split.startswith("#") and not next_split.startswith("#"):
+            better_splits.append(f"{split}\n{next_split}")
+            skip_next = True
+            continue
+
+        better_splits.append(split)
```
yoomlam, line 44: This was originally an elegant one-liner but I expanded it to be more robust to account for headings without subsequent text -- see unit test. I'm open to code improvements to make it cleaner.

ccheng26 (approved):
lgtm, I'd prefer to leave `default_chunk_splitter` as is for readability

KevinJBoyer (approved):
Forgot to add my review, LGTM!

PR #114, opened 2024-10-24
perf: Increase CPU and memory of ECS tasks

KevinJBoyer:
## Ticket

n/a

## Changes
 - Update default CPU and memory for ECS tasks such as data ingestion run in the `dev` environment

## Context for reviewers
 - The web ingestion task runs out of memory with the default of 4GB.


yoomlam (approved):
worked!

PR #116, opened 2024-10-24
perf: Use raw responses of chat context in chat history

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-560

## Changes
- add raw response (instead of HTML-formatted response) to metadata and pull from context

## Testing
add print statement to line 81, message output should be list of exchanges indicated by `role` and `content`
Screenshot 2024-10-25 at 11 16 49 AM

### DIFF
--- a/app/tests/src/test_chainlit.py
+++ b/app/tests/src/test_chainlit.py
@@ -1,11 +1,11 @@
 from unittest.mock import MagicMock
 
+import tests.src.db.models.factories as factories
 from src import chainlit, chat_engine
 from src.chainlit import _get_retrieval_metadata, get_raw_chat_history
 from src.chat_engine import OnMessageResult
 from src.db.models.document import Subsection
 from src.generate import PROMPT
-import tests.src.db.models.factories as factories
 
 
 def test_url_query_values(monkeypatch):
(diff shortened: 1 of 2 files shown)


app/tests/src/test_chainlit.py
```diff
@@ -56,3 +60,62 @@ def test__get_retrieval_metadata(chunks_with_scores):
...
+        type="user_message",
+    )
+    clMessage3.configure_mock(
+        content="<div>To be eligible for unemployment benefits, you need to have earned enough wages during a specific 12-month period called the base period (citation-46). <p>For Disability Insurance, you must have earned at least $300 in wages that were subject to State Disability Insurance (SDI) deductions (citation-123). For Paid Family Leave, you also need to have earned at least $300 with SDI deductions during your base period (citation-129).</p> If you are applying for an overpayment waiver, your average monthly income must be less than or equal to the amounts in the Family Income Level Table (citation-118).</div>",
+        metadata={
+            "system_prompt": PROMPT,
+            "chunks": [factories.ChunkFactory.build()],
+            "subsections": [
+                {
+                    "id": "citation-1",
```
yoomlam, line 97: The LLM generated response doesn't refer to `(citation-1)` and in fact, refers to citations that aren't in `subsections`, which is not typical or intended. I suggest adjusting the generated response or the `subsections`. Note the response and other `subsections[*].text` don't need to be long to test your new function.

yoomlam (approved):

app/tests/src/test_chainlit.py
ccheng26: removed chunks and subsections as they're not key values we're using

PR #118, opened 2024-10-29
feat: Add Markdown tree support utilities for hierarchical chunking

yoomlam:
## Ticket

First PR for  https://navalabs.atlassian.net/browse/DST-558

## Changes

- Add data structures and utilities for working with markdown text as a tree.
- Add Python notebook to demonstrate its use

## Testing

Run through Python notebook
See unit test


app/tests/src/ingestion/test_markdown_tree.py
```diff
@@ -0,0 +1,345 @@
...
+    assert_content(_tree)
+    assert len(tokens_vs_tree_mismatches(_tree)) == 0
+
+
+@pytest.fixture
+def tree(markdown_text):
+    return create_markdown_tree(markdown_text)
+
+
+def test_markdown_tokens_as_json(markdown_text):
```
KevinJBoyer, line 179: Quick note while I review the rest, you may have meant to move this to `test_markdown_tree.py`?

app/src/ingestion/markdown_tree.py
```diff
@@ -0,0 +1,602 @@
...
+    """
+    Render the node and its descendants (a subtree) to markdown text.
+    Useful for creating markdown text for chunks.
+    Since the structure of the tree (i.e., each node's parent and children) is independent of each Token's parent and children,
+    we cannot rely on mistletoe's renderer (which is based on Tokens) to render the tree correctly. Hence, we have this function.
+    """
+    if node.data_type == "HeadingSection":  # Render the custom HeadingSection node specially
+        out_str = []
+        for c in node.children:
+            out_str.append(render_subtree_as_md(c, normalize=normalize))
```
yoomlam, line 198: Alternatively, we could customize or create a subclass of MarkdownRenderer (https://github.com/miyuchina/mistletoe/blob/master/dev-guide.md#a-new-renderer) to handle `HeadingSection` but this is easier and more transparent, esp. since we don't need a token to represent HeadingSections.

KevinJBoyer (approved):
I'm mulling over future refactors to try to tease out a few different concepts: 1- the tree data structure functionality (nutree does some of this, but there are also functions in here that aren't MD specific), 2- transforming the Markdown AST into our preferred structure (e.g., hierarchical headings), and 3- ingestion-specific logic for our use-case (e.g., putting intro text before list items, etc.). In the meantime though, between playing around with it and the tests, I'm convinced that this works as intended, and it's a significant leap forward for functionality, so approving with the caveat that if we have time in the future I might take a stab at separating those out -- in the meantime I don't want to unnecessarily delay this from going into main. Thanks for putting this together and presenting the approach yesterday!

app/src/ingestion/markdown_tree.py
```diff
@@ -0,0 +1,580 @@
...
+    Returns a tree reflecting the structure of the Tokens parsed from the markdown text.
+    The tree is created using mistletoe's Tokens and the TokenNodeData class.
+    Note that the structure of the tree (i.e., each node's parent and children) is independent of each Token's parent and children.
+    To render the tree to markdown text, use render_tree_as_md() or render_subtree_as_md().
+    """
+    if normalize_md:
+        markdown = normalize_markdown(markdown)
+    with _new_md_renderer():
+        # Never call Document(...) outside of a with ... as renderer block"
+        # Otherwise, markdown_renderer.BlankLine will not be created
```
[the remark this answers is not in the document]
yoomlam: From https://github.com/miyuchina/mistletoe?tab=readme-ov-file#usage-from-python:
> **Important**: As can be seen from the example above, the parsing phase is currently tightly connected with initiation and closing of a renderer. Therefore, you should never call `Document(...)` outside of a `with ... as renderer` block, unless you know what you are doing.

One implication I noticed was that `BlankLine` tokens were not created, so when the set of tokens are rendered, the markdown was missing blank lines that would typically delineate block elements, i.e., consecutive paragraphs became a single paragraph.

PR #120, opened 2024-10-30
feat: Add support for retrieval in other languages

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-553

## Changes
 - Run user messages through an analyzer to understand their original language and if they need to have sources retrieved
 - Don't show citation text in the UI if they're not matched -- instead, log an error. 

## Context for reviewers
 - Translating into English for retrieval improves the sources retrieved
 - Follow up questions ("please simplify that", "please translate into simplified Chinese") often don't need retrieved sources at all, and this can in fact complicate the response by providing the LLM irrelevant text
 - References to prior messages sometimes carry along the citation references, but these won't match, so don't display them.

## Testing
This functionality involves LLMs -- open to suggestions on adding automated coverage for this functionality.

I did extensive testing locally. This screenshot captures the core of the functionality -- retrieving appropriate English sources for a question asked in Spanish; not retrieving sources when there is a follow-up question that does not require them.
image


app/src/generate.py
```diff
@@ -18,16 +20,29 @@
...
+- They may qualify if they were laid off due to lack of work.(citation-1) (citation-2)
+- They might be eligible if their hours were significantly reduced.(citation-3)
+"""
+
+ANALYZE_MESSAGE_PROMPT = """
+Analyze the user's message to determine how to respond.
+Reply with a JSON dictionary.
+Set original_language to the language of the user's message.
+If the user's message is in English, set is_in_english to true.
+Otherwise, set is_in_english to false and set message_in_english with the translation of the query into English.
```
yoomlam, line 43:
```suggestion
Otherwise, set is_in_english to false and set message_in_english to a translation of the query into English.
```

app/src/generate.py
```diff
@@ -18,16 +20,29 @@
...
+If the question would be easier to answer with additional policy or program context (such as policy documentation), set needs_context bool to True.
```
yoomlam, line 44: The `bool` type isn't mentioned earlier in the prompt so let's keep it out for consistency unless you need it.
```suggestion
If the question would be easier to answer with additional policy or program context (such as policy documentation), set needs_context to True.
```

app/src/generate.py
```diff
@@ -86,7 +102,13 @@ def generate(
         messages.extend(chat_history)
 
     messages.append({"content": query, "role": "user"})
-    logger.info("Calling %s for query: %s with context:\n%s", llm, query, context_text)
+
+    # Without this direction, OpenAI's GPT-4o will sometimes ignore the original language and respond in English.
+    messages.append(
+        {"content": f"Please translate your answer into {requested_language}", "role": "system"}
```
yoomlam, line 108: Are we still asking the LLM nicely? 😉

app/src/chat_engine.py
```diff
@@ -85,18 +85,34 @@ class BaseEngine(ChatEngineInterface):
...
+            )
+
+            chunks = [chunk_with_score.chunk for chunk_with_score in chunks_with_scores]
+            # Provide a factory to reset the citation id counter
+            subsections = split_into_subsections(chunks, factory=CitationFactory())
+            context_text = create_prompt_context(subsections)
+        else:
+            context_text = None
+            chunks_with_scores = []
+            subsections = []
```
yoomlam, line 106: `context_text` doesn't need to be set.
Consider:
```suggestion
            return OnMessageResult(response, self.system_prompt)
```
and setting defaults on `OnMessageResult` for `chunks_with_scores` and `subsections`.

yoomlam: This will also make it more obvious the 2 scenarios where `OnMessageResult` is created.

yoomlam (approved):

app/src/generate.py
KevinJBoyer: 😂  I removed the "Please". Habits!

app/src/generate.py
```diff
@@ -18,16 +20,29 @@
...
+To cite multiple sources at once, you can append citations like so: (citation-1) (citation-2), etc.
+Place the citations immediately AFTER any closing punctuation for the sentence.
+For example: 'This is a sentence that draws on information from the context.(citation-1)'
+Do NOT place the citations BEFORE the closing punctuation, or add a space between the sentence and the citation.
 
 Example Answer:
 If the client lost their job at no fault, they may be eligible for unemployment insurance benefits. For example:
-- They may qualify if they were laid off due to lack of work. (citation-1)
-- They might be eligible if their hours were significantly reduced. (citation-2)
+- They may qualify if they were laid off due to lack of work.(citation-1) (citation-2)
```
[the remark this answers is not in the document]
KevinJBoyer: Good question -- yup, I'm trying to get the superset text the LLM generates to match how you'd do this by hand

app/src/chat_engine.py
KevinJBoyer: 👍  Made this change and split the two out into separate functions as well

PR #121, opened 2024-11-01
feat: Add tree-based chunking to EDD ingestion

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-558

## Changes

* Added robust and configurable tree-based chunking algorithm
   - See `chunk_tree()`, `_gradually_chunk_tree_nodes()`, and `_handle_does_not_fit()`
   - Chunking is configured via the `ChunkingConfig` class
* Integrated with EDD ingestion

## Testing

```
cd app
DEBUG_SCRAPINGS=true poetry run scrape-edd-web
# Enable saving chunks to `chunks-log` subfolder
touch SAVE_CHUNKS
make ingest-edd-web DATASET_ID="CA EDD" BENEFIT_PROGRAM=employment BENEFIT_REGION=California FILEPATH=src/ingestion/edd_scrapings.json
```


yoomlam:
Analyzing call graph using `python-call-graph` with command `pycallgraph --exclude 'torch.*' --include 'src.*' graphviz -- src/ingestion/markdown_chunking.py` after temporarily adding:
```
if __name__ == "__main__":
    test_markdown = f"""This is the first paragraph with no heading. ..."""
    print(test_markdown)
    prepped_tree = create_markdown_tree(test_markdown)
    config = EddChunkingConfig()
    chunks = chunk_tree(prepped_tree, config)
```
image

yoomlam:
Call tree for the primary functions:
```
chunk_tree
- _gradually_chunk_tree_nodes
    - _summarize_node
    - _handle_does_not_fit
        - _add_chunks_and_summarize_node (outermost recursion)
            - _add_chunks_for_heading_section
                - _add_chunks_and_summarize_node (recurse)
                    - ...
                - _add_chunks_from_partitions
            - _add_chunks_for_list_or_table
                - _new_tree_for_partials
                - _add_chunks_and_summarize_big_listitems
                    - _add_chunks_and_summarize_node (recurse)
                        - ...
                - _add_chunks_from_partitions
            - _summarize_node
        - _summarize_node
- _add_chunks_and_summarize_node
    - ...
```

app/src/ingest_edd_web.py
```diff
@@ -86,27 +104,73 @@ def _create_chunks(
...
-        # Start a new split for each heading
-        section_splits = _split_heading_section(headings, text)
-        splits.extend(section_splits)
+    if USE_MARKDOWN_TREE:
+        chunking_config = EddChunkingConfig()
+
+        # Fix markdown formatting that causes markdown parsing errors
+        # '. . .' is parsed as sublists on the same line
+        # in https://edd.ca.gov/en/uibdg/total_and_partial_unemployment_tpu_5/
+        content = content.replace(". . .", "...")
```
ccheng26, line 141: maybe extract this into a md_formatting function

app/src/ingest_edd_web.py
```diff
@@ -86,27 +104,73 @@ def _create_chunks(
...
+
+        try:
+            tree = create_markdown_tree(content, doc_name=document.name, doc_source=document.source)
+            tree_chunks = chunk_tree(tree, chunking_config)
+
+            for chunk in tree_chunks:
+                split = SplitWithContextText(
+                    chunk.headings, chunk.markdown, chunk.context_str, chunk.embedding_str
+                )
+                assert split.token_count == chunk.length
```
KevinJBoyer, line 157: Am I following this right:
 - `chunk_tree` returns ProtoChunk
 - There is a 1:1 mapping of ProtoChunk -> SplitWithContextText
 - There is then a 1:1 mapping from SplitWithContextText -> Chunk

and from there:
 - SplitWithContextText is only used by the `if not USE_MARKDOWN_TREE` branch,
 - which we plan to deprecate in the future after adopting this chunking approach?
 
---

Then I think an approach to encapsulate the chunking logic here (ingestion/markdown_chunking.py and ingestion/markdown_tree.py) is:

1. Expose from ingestion a function `def chunk_markdown(markdown: str) -> list[Chunk]`
2. Update `ingest_edd_web.py` to only call `chunk_markdown`

That also lets us define integration-level tests that illustrate (and enforce) the behavior of the encapsulated code: we can write a set of pairs of markdown input strings, and the expected list of output Chunks, and compare the two sets.

I'm not suggesting we do that as part of this PR but I do think it'd be good to follow up: those integration-level tests would make it clearer for other engineers what the behavior of the logic is, and also make it easier to refactor the code (whether to update the behavior or otherwise.)

KevinJBoyer (approved):
Left a comment on how we might revise this code to better encapsulate the functionality here into a single public function. I'm a little torn on making that change (plus adding/revising test coverage to be at the level of that function) as part of this PR or deferring it as technical debt, so leaving that up to you.

app/src/ingest_edd_web.py
KevinJBoyer: > I do think it'd be good to follow up

(to clarify I'm not saying this needs to be you Yoom!)

yoomlam: Yes, except:

> SplitWithContextText is only used by the if not USE_MARKDOWN_TREE branch,

SplitWithContextText is used in both `if` branches:
https://github.com/navapbc/labs-decision-support-tool/pull/121/files#diff-456c2fb3cff98badccca05b82109f1bab9c91a077cdfbc21245de5db8fa69b8aR154

SplitWithContextText is the commonality between the `if` branches. `ProtoChunk` is new and only used in the tree-based chunking code.

The encapsulation idea sounds good. There's definitely more refactoring that can be done outside the tree-chunking code!

KevinJBoyer: > SplitWithContextText is used in both if branches

What I mean is that it's not actually _doing_ anything in the new `if USE_MARKDOWN_TREE` branch. The other branch calls some SplitWithContextText functionality like `SplitWithContextText.add_if_within_limit`, etc. So after we switch to the new functionality we can drop this part of the code.

yoomlam: Yes!

PR #126, opened 2024-11-14
fix: Include all headers in batch CSV generation

KevinJBoyer:
## Ticket

n/a

## Changes
 - Instead of gathering the keys from the first processed row of the CSV, gather all of the keys from all of the rows (since later rows may have more keys if they have more citations)

## Testing
Added coverage to the test for this function. The test will now fail if the change is reverted

### DIFF
--- a/app/src/batch_process.py
+++ b/app/src/batch_process.py
@@ -26,7 +26,10 @@ async def batch_process(file_path: str, engine: ChatEngineInterface) -> str:
             row.update(data)
 
         # Update fieldnames to include new columns
-        all_fieldnames = list(reader.fieldnames) + list(processed_data[0].keys())
+        all_fieldnames = list(reader.fieldnames) + [key for p in processed_data for key in p.keys()]
+        # De-duplicate while preserving order -- ignore linter warning that seen.add always returns None
+        seen = set()
+        all_fieldnames = [f for f in all_fieldnames if not (f in seen or seen.add(f))]  # type: ignore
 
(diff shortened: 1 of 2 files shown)


app/src/batch_process.py
```diff
@@ -26,7 +26,10 @@ async def batch_process(file_path: str, engine: ChatEngineInterface) -> str:
             row.update(data)
 
         # Update fieldnames to include new columns
-        all_fieldnames = list(reader.fieldnames) + list(processed_data[0].keys())
+        all_fieldnames = list(reader.fieldnames) + [key for p in processed_data for key in p.keys()]
+        # De-duplicate while preserving order -- ignore linter warning that seen.add always returns None
+        seen = set()
+        all_fieldnames = [f for f in all_fieldnames if not (f in seen or seen.add(f))]  # type: ignore
```
yoomlam, line 32: An alternative is to use a dict, which preserves order and the keys are unique. The dict values can be None.

yoomlam (approved):

app/src/batch_process.py
KevinJBoyer: @yoomlam like this?

```suggestion
        all_fieldnames = {f: None for f in list(reader.fieldnames) + [key for p in processed_data for key in p.keys()]}
        all_fieldnames = list(all_fieldnames)
```

yoomlam: Yes, the last line could be `all_fieldnames.keys()` to be explicit.

ccheng26 (approved):

PR #127, opened 2024-11-14
feat: Write document questions and answers to csv

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-420

## Changes
- Added notebook for generating question answer pairs based on document/chunks

## Testing
```
Run make notebook
navigate to http://localhost:8888/
copy and paste token in URL found in terminal
run notebook file in notebooks/question_answer_generator/q_a_generation.ipynb
question_answer_pairs.csv should be created with question, answer, document_name, document_source, and document_id fields
```


yoomlam:
Given there's only one python cell in the notebook, why was a notebook needed? In other words, why not have the code in the notebook be a `main` code block in the python file?

app/notebooks/question_answer_generator/generate_q_a_pairs.py
```diff
@@ -0,0 +1,95 @@
...
+GENERATE_QUESTION_ANSWER_PROMPT = """
+Using the provided text, generate at least 5 unique questions and answers, avoid rephrasing or changing the punctuation of the question to ensure distinct questions and answers.
+Respond with a list of JSON dictionaries in the following format  (do not wrap in JSON markers):
+question: The generated question based on the content.
+answer: The answer to the question, derived from the content.
+document_name: The name of the document.
+document_source: The source or URL of the document.
+document_id: the id of the chunk or document.
+Example
+question: What's a base period for SDI?
```
yoomlam, line 22:
```suggestion
Respond with a list of JSON dictionaries in the following format (do not wrap in JSON markers):
question: The generated question based on the content.
answer: The answer to the question, derived from the content.
document_name: The name of the document.
document_source: The source or URL of the document.
document_id: the id of the chunk or document.

Example
question: What's a base period for SDI?
```

Since the response is structured, why not have it to produce JSON?

ccheng26: I've revised the prompt several times with the term `Your response should return valid JSON` and provided some examples but the response always comes back as a string.

KevinJBoyer:
Looks great, nice use of Pydantic models to get structured JSON back out.
Two thoughts:
1 - Could we make the number of questions per chunk/document generated a user input (or just a constant like `NUM_QA_PER_CHUNK_OR_DOC = 5` in the notebook that could be easily changed)? Then I think we need to reconcile this with the two prompts in `GENERATE_QUESTION_ANSWER_PROMPT` and `process_document_or_chunk`: right now, the system prompt says "Using the provided text, generate at least 5 unique questions and answers..." and this is followed by a user message that says "Please use the following content to create {num_of_chunks} question-answer pairs." I think

2 - Could we move the model reference (`gpt-4o`) into the notebook so that it's simple for a developer to decide to use, e.g., gpt-4o-mini for a run, etc? (Could also accomplish this by moving all of the code from .py into the notebook, defer to you on that call!)

KevinJBoyer:
>Did we want to set NUM_QA_PER_CHUNK_OR_DOC as a static number in the prompt in a single place?
Yeah I think that's right!

No need to include models besides OpenAI, since that's the only one that supports response_format (AFAIK) at the moment.

app/notebooks/question_answer_generator/generate_q_a_pairs.py
```diff
@@ -0,0 +1,107 @@
...
+    question_answer_list: list[QuestionAnswerAttributes] = []
+
+    for generated_question_answer in generated_question_answers.pairs:
+        document_item = document if isinstance(document, Document) else document.document
+        question_answer_item = QuestionAnswerAttributes(
+            document_id=document_item.id,
+            document_name=document_item.name,
+            document_source=document_item.source,
+            question=generated_question_answer.question,
+            answer=generated_question_answer.answer,
```
KevinJBoyer, line 85: We need the `chunk.id` if the Q&A was generated from a chunk (so that when we later query with the question, we can see if the matching chunk.id is retrieved)

KevinJBoyer (approved):
LGTM but need to make one more change, which is saving the chunk ID (if the question and answers were generated from a specific chunk) to the CSV as well

app/notebooks/question_answer_generator/generate_q_a_pairs.py
ccheng26: `document_id` should be `chunk.id` if a chunk is passed since `document_item` can be a `Chunk` or `Document`

ccheng26: Oversight on my part, thanks for the catch! I'll make the `chunk_id` more explicit as well

PR #128, opened 2024-11-14
fix: Have superscript citation link to source URL

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-603

## Changes

Have superscript citation link to source URL

## Context for reviewers

https://nava.slack.com/archives/C06ETE82UHM/p1731513849101759

## Testing

image

PR #129, opened 2024-11-14
refactor: Use FormattingConfig in format.py

yoomlam:
## Ticket

Related to #128

## Changes

Refactor to use `FormattingConfig` and `BemFormattingConfig` instead of `if-then` blocks.

### DIFF
--- a/app/src/chat_engine.py
+++ b/app/src/chat_engine.py
@@ -171,7 +171,7 @@ class BridgesEligibilityManualEngine(BaseEngine):
 
 
 class CaEddWebEngine(BaseEngine):
-    retrieval_k: int = 5
+    retrieval_k: int = 50
     retrieval_k_min_score: float = -1
 
     # Note: currently not used
--- a/app/src/chat_engine.py
+++ b/app/src/chat_engine.py
@@ -5,7 +5,7 @@ from typing import Callable, Sequence
 
 from src.citations import CitationFactory, create_prompt_context, split_into_subsections
 from src.db.models.document import ChunkWithScore, Subsection
-from src.format import format_bem_subsections, format_guru_cards, format_web_subsections
+from src.format import BemFormattingConfig, FormattingConfig, format_guru_cards
 from src.generate import PROMPT, MessageAttributes, analyze_message, generate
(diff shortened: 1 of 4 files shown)


app/tests/src/test_format.py
```diff
@@ -264,35 +267,32 @@ def test__return_citation_link():
...
     )
 
     assert (
-        format_web_subsections(0, 0, [], [], "List intro sentence: \n- item 1\n- item 2")
+        build_accordions([], "List intro sentence: \n- item 1\n- item 2", config)
         == "<div><p>List intro sentence: </p>\n<ul>\n<li>item 1</li>\n<li>item 2</li>\n</ul></div>"
     )
 
-    chunks_with_scores[0].chunk.document.name = "Your State Disability Insurance (SDI)"
-    chunks_with_scores[1].chunk.document.name = "Your State Disability Insurance (SDI): Another"
```
yoomlam, line 292: This refactoring revealed that `chunks_with_scores` is not relevant to `format_web_subsections` (now `build_accordions`)

app/tests/src/test_format.py
```diff
@@ -264,35 +267,32 @@ def test__return_citation_link():
...
     chunk_list[1].page_number = 3
 
-    bem_link = return_citation_link(chunk_list[0], "BEM")
+    bem_link = BemFormattingConfig().return_citation_link(chunk_list[0])
 
     assert "Open document to page 3" in bem_link
     assert "Source" not in bem_link
 
-    web_link = return_citation_link(chunk_list[1], "EDD")
+    web_link = FormattingConfig().return_citation_link(chunk_list[1])
```
yoomlam, line 275: It is now clear that we only use 2 FormattingConfigs

app/src/chat_engine.py
yoomlam, line 1: Note on line 39, we're using the same LLM system prompt for all chat engines! This means all engines are expecting to talk about California and EDD.

yoomlam: #130

KevinJBoyer (approved):
LGTM, thank you!

PR #132, opened 2024-11-19
feat: Add minimal API v0

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-605

## Changes

* Enable Fast API on startup
* Add `query` API endpoint
* Add instructions to `docs/app/getting-started.md`

## Testing

* Start app make sure `/chat` still works
* Go to http://localhost:8000/api/healthcheck
* http://localhost:8000/api/engines?user_id=myName should return `["ca-edd-web"]` for any `user_id` for now.
* `curl -X POST 'http://0.0.0.0:8000/api/query' -H 'Content-Type: application/json' -d '{ "session_id": "12", "new_session": true, "message": "unemployment insurance benefit deadlines?"}' | jq` should return a valid response
* `curl -v -X POST 'http://0.0.0.0:8000/api/query' -H 'Content-Type: application/json' -d '{ "session_id": "12", "new_session": true}'` should return an error `HTTP/1.1 422 Unprocessable Entity`

Go to https://cloud.getliteral.ai/ and check the threads:
image
image

* Go to http://localhost:8000/docs for API docs


app/Makefile
```diff
@@ -106,6 +106,10 @@ stop:
 
 check: format-check lint test
 
+launch-api-only:
+	# Only for local development
+	poetry run launch-api-only -p 8001
```
KevinJBoyer, line 111:
```suggestion
	$(PY_RUN_CMD) launch-api-only -p 8001
```

This automatically switches between local and in docker depending on the dev's PY_RUN_APPROACH setting

app/src/chat_api.py
```diff
@@ -0,0 +1,240 @@
...
+async def healthcheck(request: Request) -> HealthCheck:
+    logger.info(request.headers)
+    healthcheck_response = await health(request)
+    return healthcheck_response
+
+
+@functools.cache
+def literalai() -> AsyncLiteralClient:
+    """
+    This needs to be a function so that it's not immediated instantiated upon
```
KevinJBoyer, line 38:
```suggestion
    This needs to be a function so that it's not immediately instantiated upon
```

app/src/chat_api.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def get_chat_engine(session: UserSession) -> ChatEngineInterface:
+    engine_id = session.chat_engine_settings.engine_id
+    # May want to cache engine instances rather than creating them for each request
+    engine = (
+        chat_engine.create_engine(engine_id) if engine_id in session.user.allowed_engines else None
+    )
+    if not engine:
+        raise HTTPException(status_code=406, detail=f"Unknown engine: {engine_id}")
```
KevinJBoyer, line 168: strictly speaking I think this should be 401 (though TIL about 406!)

app/src/chat_api.py
```diff
@@ -0,0 +1,240 @@
...
+
+def main() -> None:  # pragma: no cover
+    import getopt
+    import sys
+
+    import uvicorn
+    from fastapi import FastAPI
+
+    # Use default port 8001 so that this standalone API app does not conflict with the Chainlit app
+    port = 8001
```
KevinJBoyer, line 230: What's the reason to run this on a different port when the API is invoked standalone? (And as a result, if I'm following right, only support standalone mode locally and not in the Docker container?)

KevinJBoyer (approved):
LGTM, with open question about the standalone configuration

app/Makefile
yoomlam: I changed it from `$(PY_RUN_CMD)` because I'd have create a docker config to make port 8001 available outside the Docker environment. As the comment says this is `Only for local development`.

app/src/chat_api.py
yoomlam: I intend this for `Only for local development` currently. If/once we decide to run the API without the Chatbot web UI, then we can make this a more permanent change and use the application port 8000.

KevinJBoyer: Right, my question is why -- it seems like if you just leave it on port 8000 it'll work in Docker too, right?

app/src/chat_api.py
yoomlam: 401 sounds like an authentication problem, which this is not. How about 403?
> similar to [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401), except that for 403 Forbidden responses, authenticating or re-authenticating makes no difference. The request failure is tied to application logic, such as insufficient permissions to a resource or action.

yoomlam, line 168:
```suggestion
        raise HTTPException(status_code=403, detail=f"Unknown engine: {engine_id}")
```

app/src/chat_api.py
KevinJBoyer: Yes, I think we should continue to support both methods of running the app locally: either directly on the machine (poetry run...) or in a Docker container.

PR #133, opened 2024-11-19
refactor: Add ChatHistory class

yoomlam:
## Ticket

Associated with https://navalabs.atlassian.net/browse/DST-605

## Changes

refactor: Add ChatHistory class

## Testing

No behavior change

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -12,7 +12,7 @@ from src.app_config import app_config
 from src.batch_process import batch_process
 from src.chat_engine import ChatEngineInterface, OnMessageResult
 from src.format import build_accordions
-from src.generate import get_models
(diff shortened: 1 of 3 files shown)


app/src/chainlit.py
```diff
@@ -4,6 +4,8 @@
 from typing import Any
 from urllib.parse import parse_qs, urlparse
 
+from asyncer import asyncify
```
KevinJBoyer, line 7: Just curious, any particular reason to prefer this over the cl.make_async func?

KevinJBoyer (approved):

app/src/chainlit.py
yoomlam: So that the piece of code is not chainlit-specific and the line can later be refactored out of chainlit.py

PR #142, opened 2024-11-21
feat: Add chat_message DB table

yoomlam:
## Ticket

Part 2 for https://navalabs.atlassian.net/browse/DST-606

## Changes

* Add chat_message DB table
* Update chat_api.py to use DB

## Testing

Perform same test in #140


app/src/chat_api.py
```diff
@@ -226,8 +218,8 @@ async def query(request: QueryRequest) -> QueryResponse:
                 detail=f"Chat history for existing session not found: {request.session_id}",
             )
 
-        _save_message(
-            ChatMessage(request.session_id, "user", request.message, request_msg.timestamp)
+        db_session.add(
+            ChatMessage(session_id=request.session_id, role="user", content=request.message)
```
ccheng26, line 222: non-blocking: should we have an enum for the roles "user" and "assistant" defined somewhere?

yoomlam: I considered that, but since we're prototyping/piloting, I left it open in case we want to experiment with other roles.

ccheng26 (approved):

PR #143, opened 2024-11-26
feat: Add Imagine LA engine

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-586

## Changes
 - Add new chat engine that pulls from both Imagine LA and EDD web datasets
 - Set as the default both for web and API interfaces

## Context for reviewers
 - This will be the engine we use for the pilot going forward
 - I'll seek input from design on how we can update the prompt here, I edited it to be generic to CA (rather than EDD specific) for now

## Testing
image

### DIFF
--- a/app/src/chat_engine.py
+++ b/app/src/chat_engine.py
@@ -202,3 +202,22 @@ If you can't find information about the user's prompt in your context, don't ans
 If a prompt is about an EDD program, but you can't tell which one, detect and clarify program ambiguity. Ask: "The EDD administers several programs such as State Disability Insurance (SDI), Paid Family Leave (PFL), and Unemployment Insurance (UI). I'm not sure which benefit program your prompt is about; could you let me know?"
 
 {PROMPT}"""
+
+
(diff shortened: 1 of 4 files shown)


app/src/chat_engine.py
```diff
@@ -202,3 +202,22 @@ class CaEddWebEngine(BaseEngine):
...
+
+    # Note: currently not used
+    chunks_shown_min_score: float = -1
+    chunks_shown_max_num: int = 8
+
+    engine_id: str = "imagine-la"
+    name: str = "Imagine LA Chat Engine"
+    datasets = ["CA EDD", "Imagine LA"]
+
+    system_prompt = f"""You are an assistant to navigators who support clients-such as claimants, beneficiaries, families, and individuals-during the screening, application, and receipt of public benefits in California.
```
ccheng26, line 219: maybe use `()` instead of `-`

ccheng26 (approved):
code lgtm

PR #145, opened 2024-12-02
fix: Remove use of getopt for ingestion

yoomlam:
## Ticket

Related to https://navalabs.atlassian.net/browse/DST-564

## Changes

Remove use of `getopt` for ingestion since we're not using any [getopt options](https://docs.python.org/3/library/getopt.html) -- we're using "arguments" only.

## Testing

No behavior change. 
Run some ingestion command.

### DIFF
--- a/app/src/util/ingest_utils.py
+++ b/app/src/util/ingest_utils.py
@@ -1,4 +1,3 @@
-import getopt
 import json
 import logging
 import re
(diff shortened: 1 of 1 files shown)


app/src/util/ingest_utils.py
```diff
@@ -36,10 +35,7 @@ def process_and_ingest_sys_args(argv: list[str], logger: Logger, ingestion_call:
         )
         return
 
-    _, args = getopt.getopt(
-        argv[1:], shortopts="", longopts=["DATASET_ID BENEFIT_PROGRAM BENEFIT_REGION FILEPATH)"]
```
yoomlam, line 40: There's also a typo: `FILEPATH)`

ccheng26 (approved):

PR #149, opened 2024-12-05
feat: Tree-based chunk splitting into subsections

yoomlam:
## Ticket

Follow-on to #148 for https://navalabs.atlassian.net/browse/DST-627
for robustness, BUT merging this is not necessary IF all `chunk.content` is "correctly" formatted, which it should be if we continue to use the tree-based chunking, which has the markdown normalization step to correctly format markdown.

## Changes

* More robust splitting of chunks. Prior implementation relies on `\n\n`. This implementation parses the markdown structure and can handle acceptable markdown text variations like:
```
## Some Heading
Text with no empty line between heading and paragraph.
```

## Testing

Unit test


app/src/citations.py
```diff
@@ -33,28 +36,67 @@ def create_citation(self, chunk: Chunk, text: str, text_headings: Sequence[str])
...
     chunk: Chunk, factory: CitationFactory = citation_factory
+) -> list[Subsection]:
+    try:
+        return tree_based_chunk_splitter(chunk, factory)
+    except RuntimeError as e:
+        logger.warning(
+            "Falling back to basic_chunk_splitter for chunk: %s", chunk.id, exc_info=True
+        )
+        logger.warning(e)
+        return basic_chunk_splitter(chunk, factory)
```
yoomlam, line 47: Since this is run during user interaction, fall back to the basic splitter and log a warning to diagnose the error.

KevinJBoyer (approved):

PR #150, opened 2024-12-05
feat: Analyze precision and recall for initial eval dataset

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-592

## Changes
 - Move Q&A generation into a `metrics` subfolder; modify script to save the content_hash (md5) to the generated CSV
 - Add another notebook for computing precision/recall on CSVs of eval questions

## Context for reviewers
 - Rather than commit the eval dataset to the repo, I've uploaded it to a Google Sheet [here](https://docs.google.com/spreadsheets/d/1KBFMyRUSohqA94ic6yAv3Ne22GwEBJHHYHM49rEKFsc/edit?usp=sharing) (viewable only by the Nava team)
 - Although I initially planned to add these as automated tests that could be manually invoked, I ended up including the code in a notebook instead, to support easy interaction (since this is more about analysis than asserting facts about the codebase). If folks feel strongly about instead going with the original approach of adding them as pytest tests that are skipped by default, I can revert back.

## Testing
On the provided Google sheet, with questions generated from the Imagine LA content hub:
Screenshot 2024-12-05 at 9 40 03 AM
Screenshot 2024-12-05 at 9 40 07 AM
(Arbitrarily low K chosen to show interesting values for both.)


app/notebooks/metrics/q_a_generation.ipynb
```diff
@@ -0,0 +1,624 @@
+{
```
KevinJBoyer, line 1: for reviewers: this file was slightly modified, but git didn't pick up that it was moved, show it shows up as a new file with the old one deleted.

app/notebooks/metrics/generate_q_a_pairs.py
```diff
@@ -70,28 +72,29 @@ def generate_question_answer_pairs(llm: str, message: str) -> QuestionAnswerList
...
-            document_id=document_item.id,
-            document_name=document_item.name,
-            document_source=document_item.source,
+            document_id=document.id,
+            document_name=document.name,
+            document_source=document.source,
             question=generated_question_answer.question,
             answer=generated_question_answer.answer,
-            chunk_id=None if is_document else document.id,
+            chunk_id=None if is_document else document_or_chunk.id,
```
yoomlam, line 96: I'd refactor this further for improved readability to something like:
```
if isinstance(document_or_chunk, Document):
  document = document_or_chunk
  chunk_id = None
else:
  document = document_or_chunk.document
  chunk_id = document_or_chunk.id
```

yoomlam (changes requested):
For the CSV file, should probably add quotes around the questions in case they have commas.

I got `NameError: name 'widgets' is not defined` for `PrecisionRecall.ipynb` line 7.

app/notebooks/metrics/generate_q_a_pairs.py
KevinJBoyer: nice, much better!

KevinJBoyer:
> For the CSV file, should probably add quotes around the questions in case they have commas.

This is handled automatically, the Python csv writer adds them when they're needed (you can see the behavior in the `answer` column, since none of the generated questions happen to have commas at the moment.)

> I got NameError: name 'widgets' is not defined for PrecisionRecall.ipynb line 7.

Should be fixed now -- I dropped an import by accident but didn't notice because the kernel had already loaded it.

yoomlam (approved):

PR #151, opened 2024-12-06
feat: Literal ai feedback

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-614

## Changes

> Added endpoint `feedback` to accept  `session_id`, `is_positive`, `response_id`, `comment` and `user_id`, send the values to literalai, and return a response value feedback created successfully

## Testing
1. run `make start`
2. query the api via localhost:8000/api/query using similar payload to the following:
```
{
  "session_id": "your session name",
  "new_session": "true",
  "message": "What is a base period for SDI?" 
}
```

3. make note of the `response_id`
4. call the feedback endpoint localhost:8000/api/feedback with the payload:

```
{
    "session_id": "your session name",
    "is_positive": "true",
    "response_id": response_id from previous call,
    "comment": "great"
}
```

5. results should be reflected in LiteralAI logs; if `is_positive` is "true", the score value should be 1, if it's "false" the score value should be 0
Screenshot 2024-12-06 at 5 04 50 PM

Screenshot 2024-12-06 at 5 05 19 PM


app/tests/src/test_chat_api.py
```diff
@@ -202,3 +232,42 @@ def test_get_chat_engine_not_allowed(user_info):
     )
     with pytest.raises(HTTPException, match="Unknown engine: bridges-eligibility-manual"):
         get_chat_engine(session)
+
+
+@pytest.mark.asyncio
```
yoomlam, line 237: I don't think this is needed:
```suggestion
```

docs/app/api_spec.md
```diff
@@ -102,10 +105,81 @@ Note: we will defer making any of these required for now
...
+### Request Parameters
+
+**Headers**
+
+- **Authorization**: Bearer token for authenticating API access.
+
+**Parameters**
+
+- **user_id** _(string, optional)_: Unique identifier for the navigator within ImagineLA.
+- **comment** _(string, optional)_: The user's feedback comment for the LLM response.
```
yoomlam, line 123: Move optional parameters after the required ones.

KevinJBoyer (approved):
LGTM, nice work!

docs/app/api_spec.md
```diff
@@ -102,10 +105,81 @@ Note: we will defer making any of these required for now
...
+
+- **session_id** _(string, required)_: Unique identifier for the current session, to track prior messages in the conversation.
+- **response_id** _(string, required)_: Unique identifier for the chatbot response.
+- **is_positive** _(bool, required)_: If the chatbot response is helpful to the navigator or not.
```
yoomlam, line 127: Since LiteralAI logs this as a `float`, wondering if we should make this a float. @KevinJBoyer thoughts?

docs/app/api_spec.md
```diff
@@ -102,10 +105,81 @@ Note: we will defer making any of these required for now
...
+#### Success Response (200 OK)
+
+**Body**
+
+```json
+{
+  "user_id": "string",       // Unique identifier for the navigator within ImagineLA
+  "value": "float",          // 1 if is_positive was "true" and 0 if is_positive was "false"
+  "step_id": "string",       // ID of the step associated with the score
+  "comment": "string"        // Navigator's comment on the response
```
yoomlam, line 142: Why are these sent back to the client? They're not used for anything, right?

docs/app/api_spec.md
KevinJBoyer: I'd keep it as a bool since we only plan to support thumbs up/ thumbs down. If the design requirements change in the future (and we decide to let users respond with some continuous value) then we can update the API.

docs/app/api_spec.md
KevinJBoyer: FWIW I also wondered about this -- `step_id` seems to always match `response_id`. Could also just return an empty 200 OK response to verify the feedback was received.

docs/app/api_spec.md
```diff
@@ -102,10 +105,81 @@ Note: we will defer making any of these required for now
...
+
+### Description
+
+This endpoint accepts feedback from a navigator and responds with the user_id, value, step_id and comment given for the feedback request.
+
+### Request Parameters
+
+**Headers**
+
+- **Authorization**: Bearer token for authenticating API access.
```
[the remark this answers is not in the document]
KevinJBoyer: 👁️ ah I missed that it's there as well. Yes, please remove it in this PR!

PR #152, opened 2024-12-09
fix: Handle table markdown formatting in assertion

yoomlam:
## Ticket

[Slack](https://nava.slack.com/archives/C06ETE82UHM/p1733771384506999?thread_ts=1733771029.694449&cid=C06ETE82UHM)

## Changes

- Ignore dashes (for table markdown text) as well as spaces when checking if subsection text is in the chunk text.
- Also downgrade to a `log.warning` (rather than `assert`) since this check occurs during user interaction

## Testing

Added unit test


KevinJBoyer (approved):
LGTM, thanks for including the unittest!

PR #155, opened 2024-12-11
feat: Citation footnotes open accordions

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-628

## Changes
- added matching function for accordion ids to citation footnotes
- add listener for DOM changes and add onclick events to accordions through custom js file

## Context for reviewers

> adding functions to python files results in an error with the frontend interpreting them as string values

## Testing
1. Navigate to localhost:8000
2. Ask chatbot query
3. click on citation footnote, accordion should open
citation

### DIFF
--- a/app/public/custom_script.js
+++ b/app/public/custom_script.js
@@ -23,7 +23,7 @@ let observer = new MutationObserver(function (mutations, observer) {
     }
 });
 
-observer.observe(document.body, {
+observer.observe(document.getElementById("root"), {
     subtree: true,
     attributes: true
 });
--- a/app/public/custom_script.js
+++ b/app/public/custom_script.js
@@ -23,8 +23,7 @@ let observer = new MutationObserver(function (mutations, observer) {
     }
 });
 
-let steps = document.getElementsByClassName("step")
-observer.observe(steps, {
+observer.observe(document.body, {
     subtree: true,
     attributes: true
 });
--- a/app/public/custom_script.js
+++ b/app/public/custom_script.js
@@ -1,0 +1,29 @@
+const addScriptToHeader = (src) => {
+    const script = document.createElement('script');
+    script.src = src;
+    script.async = true; // Optional: Load the script asynchronously
+
+    document.head.appendChild(script);
+}
+
+addScriptToHeader("https://cdnjs.cloudflare.com/ajax/libs/uswds/3.8.1/js/uswds.min.js")
+
+MutationObserver = window.MutationObserver || window.WebKitMutationObserver;
+
+let observer = new MutationObserver(function (mutations, observer) {
+    let accordionItems = document.getElementsByClassName("accordion_item")
+    for (let i = 0; i < accordionItems.length; i++) {
+        accordionItems[i].addEventListener("click", function () {
+            const data_id = this.dataset.id
+            const accordionButton = document.querySelector(`[aria-controls="${data_id}"]`)
+            accordionButton.setAttribute("aria-expanded", "true");
+            const accordionBody = document.getElementById(data_id)
+            accordionBody.removeAttribute("hidden");
+        });
+    }
+});
+
+observer.observe(document.body, {
+    subtree: true,
+    attributes: true
+});
--- a/app/public/custom_script.js
+++ b/app/public/custom_script.js
@@ -23,7 +23,8 @@ let observer = new MutationObserver(function (mutations, observer) {
     }
 });
 
-observer.observe(document.body, {
+let steps = document.getElementsByClassName("step")
+observer.observe(steps, {
     subtree: true,
     attributes: true
 });
--- a/app/src/format.py
+++ b/app/src/format.py
@@ -388,14 +388,6 @@ def _add_citation_links(
     footnote_list = []
 
     # Replace (citation-<index>) with the appropriate citation
-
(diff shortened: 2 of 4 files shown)


app/src/format.py
```diff
@@ -387,7 +409,8 @@ def replace_citation(match: Match) -> str:
 
         chunk = remapped_citations[citation_id].chunk
         link = config.get_superscript_link(chunk)
-        citation = f"<sup><a href={link!r}>{remapped_citations[citation_id].id}</a>&nbsp;</sup>"
+        matched_accordion_num = find_accordion_id(remapped_citations[citation_id].id)
+        citation = f"<sup><a class='accordion_item' data-id=a-{matched_accordion_num}>{remapped_citations[citation_id].id}</a>&nbsp;</sup>"
```
KevinJBoyer, line 413: Should we put that in quotes, like `data-id='a-{matched_accordion_num}'`?

app/tests/src/test_format.py
```diff
@@ -186,15 +186,19 @@ def test_reify_citations():
...
 
     assert (
         reify_citations(
             f"This is a citation ({subsections[0].id}) and another ({subsections[1].id}).",
             subsections,
             config,
+            None,
         )
-        == "This is a citation <sup><a href='#'>1</a>&nbsp;</sup> and another <sup><a href='#'>2</a>&nbsp;</sup>."
+        == "This is a citation <sup><a class='accordion_item' data-id=a-None>1</a>&nbsp;</sup> and another <sup><a class='accordion_item' data-id=a-None>2</a>&nbsp;</sup>."
```
KevinJBoyer, line 201: Could you modify this test so that `find_accordion_id` is tested (i.e., the data-ids are not None)?

app/src/format.py
```diff
@@ -134,11 +134,15 @@ def build_accordions(
...
     citations_html = ""
+    list_of_accordion_ids = []
     for document, cited_subsections in _group_by_document(remapped_citations).items():
         _accordion_id += 1
         citation_body = _build_citation_body(config, document, cited_subsections)
         formatted_citation_body = config.format_accordion_body(citation_body)
         citation_numbers = [citation.id for citation in cited_subsections]
+        list_of_accordion_ids.append(
+            {"citation_numbers": citation_numbers, "accordion_id": _accordion_id}
+        )
```
KevinJBoyer, line 145: Instead of a list, could this be a dictionary where the accordion_id was the key, and the values where citation_numbers?

app/src/format.py
```diff
@@ -369,13 +380,24 @@ def reify_citations(
 
 # FIXME: Refactor to reduce code replication with replace_citation_ids()
 def _add_citation_links(
-    response: str, remapped_citations: dict[str, Subsection], config: FormattingConfig
+    response: str,
+    remapped_citations: dict[str, Subsection],
+    config: FormattingConfig,
+    list_of_accordion_ids: list[dict] | None,
```
KevinJBoyer, line 386: I don't think you need to accept `None` here, you could require an empty list (or dictionary)

app/public/custom_script.js
```diff
@@ -0,0 +1,29 @@
...
+            const data_id = this.dataset.id
+            const accordionButton = document.querySelector(`[aria-controls="${data_id}"]`)
+            accordionButton.setAttribute("aria-expanded", "true");
+            const accordionBody = document.getElementById(data_id)
+            accordionBody.removeAttribute("hidden");
+        });
+    }
+});
+
+observer.observe(document.body, {
```
KevinJBoyer, line 26: Is there a way to narrow this further to just the chat window where we expect the accordions to be generated?

KevinJBoyer (commented):
Minor nit -- the cursor when put over links is the writing cursor (`text`) rather than `pointer` -- is it possible to fix that? (I think this is new behavior, though I'm not sure what change here introduced it)

app/public/custom_script.js
ccheng26: I've adjusted the element monitoring to root which is one level below the doc body. Unfortunately the DOM tree looks something like this:
Screenshot 2024-12-16 at 5 03 57 PM
There's multiple divs under the `MuiStack-root` class where the new text should appear. We could use the `IntersectionObserver` api, but I don't see too much of a benefit to iterating over the classes and observing, over monitoring at the root level.

app/src/format.py
```diff
@@ -369,13 +376,24 @@ def reify_citations(
...
     footnote_list = []
 
     # Replace (citation-<index>) with the appropriate citation
+
+    def find_accordion_id(citation_num: str) -> str | None:
+        if map_of_accordion_ids:
+            for key, value in map_of_accordion_ids.items():
+                if citation_num in value:
+                    return key
+        return None
```
KevinJBoyer, line 395: I think you should flip the key/value of map_of_accordion_ids: each key should be a citation number, and the value under that key should be the accordion ID. That makes this function
```
def find_accordion_id(citation_num: str) -> str | None:
    return map_of_accordion_ids[citation_num] if citation_num in map_of_accordion_ids else None
```
 which is small enough that you could probably do it directly on line 408

KevinJBoyer (approved):
One suggestion around simplifying the code by inverting the key/value dichotomy of the map_of_accordion_ids structure, other changes LGTM!

PR #162, opened 2024-12-19
feat: Provide direct link to text

ccheng26:
## Ticket

https://navalabs.atlassian.net/browse/DST-565

## Changes

> Added uri formatter to return highlighted text when navigating to html site

## Context for reviewers

> This won't work for the socialbenefitsnavigator site. The accordions hide the text content and the direct links don't work unless you expand the values and refresh

## Testing

1. run make start
2. query the api via localhost:8000/api/query using similar payload to the following:
{
  "session_id": "your session name",
  "new_session": "true",
  "message": "What programs are available to me as a low income parent?" 
}
3. find a `https://edd.ca.gov/en/` link, `socialbenefitsnavigator` links have all answers in accordions, with the first link there may be text outside of an accordion to view highlighted text


app/src/chat_api.py
```diff
@@ -197,15 +198,33 @@ class Citation(BaseModel):
...
+        citation_without_special_chars = re.sub(r"\W+", " ", subsection_text).strip()
+        citation_arr = citation_without_special_chars.split(" ")
+        formatted_text_to_highlight = (
+            "%20".join(citation_arr[:5]) + "," + "%20".join(citation_arr[-5:])
+        )
+        return (
+            (source_url[:-1] if source_url.endswith("/") else source_url)
+            + "#:~:text="
+            + formatted_text_to_highlight
+        )
```
KevinJBoyer, line 214: I might pull this out into a free-standing function since it's not specific to the Citation class (e.g., this could go in a generic utilities file)

KevinJBoyer (approved):
LGTM, thank you!

PR #165, opened 2024-12-19
bug: set message_in_english key to optional

ccheng26:
## Ticket

N/A

Screenshot 2024-12-19 at 4 10 31 PM

## Testing

1. open chatbot at localhost:8000/chat
2. enter query `what tax credits are available to families with young children`
3. should not return error above


KevinJBoyer:
Going to close without merging, but will separately look into the mock sentence transformer implementation to see if we have a bug lurking there

PR #169, opened 2025-01-09
fix: Switch to sequential processing in batch_process to resolve thread-safety issues

fg-nava:
# fix: Switch to sequential processing in batch_process to resolve thread-safety issues

## Ticket

https://navalabs.atlassian.net/browse/DST-688

## Changes

- Modified `batch_process.py` to use sequential processing instead of parallel processing
- Added explanatory comments about the thread-safety concerns with LiteLLM

## Context for reviewers

This PR addresses the Docker deployment crashes (CPU >1400%) issue when processing CSVs with multiple rows. Investigation revealed that the high CPU usage was caused by thread-safety issues in the underlying LiteLLM client libraries when running in parallel.

The fix is simple but effective: we've removed the parallel processing implementation and switched to sequential processing. This change resolves the CPU spike issues we were seeing in Docker deployments.

### Follow-up Items (to be tracked in separate tickets):
1. UI feedback issue: "File processed, results attached" message appears in logs but not in UI
2. Assessment needed: Determine if parallel processing is necessary for performance and if so, investigate thread-safe alternatives
3. Test coverage: Add integration tests for batch runs of `_process_question` (currently only have mock calls to chat engine)

## Testing

Tested locally by:
- Runnin processing with multi-row CSVs
- Confirmed CPU usage remains stable
- Verified successful processing of all rows in sequential order

The change eliminates the >1400% CPU spikes previously observed in Docker deployments while maintaining functionality.

### DIFF
--- a/app/src/batch_process.py
+++ b/app/src/batch_process.py
@@ -16,9 +16,13 @@ async def batch_process(file_path: str, engine: ChatEngineInterface) -> str:
         rows = list(reader)  # Convert reader to list to preserve order
         questions = [row["question"] for row in rows]
 
-        # Process questions in parallel while preserving order
-        with ThreadPoolExecutor() as executor:
-            processed_data = list(executor.map(lambda q: _process_question(q, engine), questions))
+        # Process questions sequentially to avoid thread-safety issues with LiteLLM
+        # Previous parallel implementation caused high CPU usage due to potential thread-safety
+        # concerns in the underlying LLM client libraries
+        processed_data = []
+        for q in questions:
+            processed_data.append(_process_question(q, engine))
+
 
         # Update rows with processed data while preserving original order
         for row, data in zip(rows, processed_data, strict=True):
--- a/app/src/batch_process.py
+++ b/app/src/batch_process.py
@@ -1,6 +1,5 @@
 import csv
 import tempfile
-from concurrent.futures import ThreadPoolExecutor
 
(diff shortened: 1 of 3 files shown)


app/src/batch_process.py
```diff
@@ -16,9 +15,12 @@ async def batch_process(file_path: str, engine: ChatEngineInterface) -> str:
...
 
-        # Process questions in parallel while preserving order
-        with ThreadPoolExecutor() as executor:
-            processed_data = list(executor.map(lambda q: _process_question(q, engine), questions))
+        # Process questions sequentially to avoid thread-safety issues with LiteLLM
+        # Previous parallel implementation caused high CPU usage due to potential thread-safety
+        # concerns in the underlying LLM client libraries
+        processed_data = []
+        for q in questions:
+            processed_data.append(_process_question(q, engine))
```
yoomlam, line 23:
```suggestion
        processed_data = [_process_question(q, engine) for q in questions]
```

yoomlam (approved):

fg-nava:
This can be closed without merging as it is upstream of #170

yoomlam:
@fg-nava You should have merged this PR so that the title "fix: Switch to sequential processing in batch_process" is included as a separate commit message in `main`. Then merge #170, which will have a separate commit message that says "fix: Ensure UI feedback message for batch processing completion"

PR #170, opened 2025-01-09
fix: Ensure UI feedback message for batch processing completion

fg-nava:
# fix: resolve missing batch processing UI feedback message

## Ticket

https://navalabs.atlassian.net/browse/DST-709

## Changes

- Updated message formatting in `chainlit.py` to ensure completion status is visible in UI
- Added proper metadata tracking for processing status
- Made completion message more descriptive by including result filename

## Context for reviewers

The batch processing feature was not showing completion messages in the UI, even though they were being logged successfully. Investigation showed that messages with author "backend" were not consistently visible in the Chainlit UI.

Key changes:
- Added metadata to include processing status and file information
- Made completion message more informative by including the output filename
- Ensured consistent message visibility throughout the processing flow

This PR is part of improving the batch processing feature's user experience, following the fix for CPU usage issues in DST-688.

## Testing

Tested locally with the following scenarios:
1. Uploaded a CSV file with valid questions
   - Confirmed "File processed, results attached" message appears with download link
   - Verified file downloads successfully
2. Uploaded a CSV without a "question" column
   - Confirmed error message appears with proper formatting

3. Verified all status messages appear in sequence:
   - "Please upload a CSV file..."
   - "Received file, processing..."
   - "File processed, results attached"
   
4. Screenshot:

   Screenshot 2025-01-09 at 1 31 45 PM

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -260,7 +260,6 @@ async def _batch_proccessing(file: AskFileResponse) -> None:
         # E.g., "abcd.csv" to "abcd_results.csv"
         result_file_name = file.name.removesuffix(".csv") + "_results.csv"
 
-        elements = [cl.File(name=result_file_name, path=result_file_path)]
         await cl.Message(
             author="backend",
             content="File processed, results attached.",
@@ -269,7 +268,7 @@ async def _batch_proccessing(file: AskFileResponse) -> None:
                 "original_file": file.name,
                 "result_file": result_file_name
             },
-            elements=elements
+            elements=[cl.File(name=result_file_name, path=result_file_path)]
         ).send()
 
     except ValueError as err:
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -213,6 +213,7 @@ async def on_message(message: cl.Message) -> None:
             metadata=_get_retrieval_metadata(result),
         ).send()
     except Exception as err:  # pylint: disable=broad-exception-caught
+        logger.exception("Error processing message: %r", message.content)
         await cl.Message(
             author="backend",
             metadata={"error_class": err.__class__.__name__, "error": str(err)},
@@ -261,6 +262,7 @@ async def _batch_proccessing(file: AskFileResponse) -> None:
         ).send()
 
     except ValueError as err:
+        logger.error("Error processing file %r: %s", file.name, err)
         await cl.Message(
             author="backend",
             content=f"Error processing file: {err}",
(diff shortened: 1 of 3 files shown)


app/src/chainlit.py
```diff
@@ -260,14 +260,25 @@ async def _batch_proccessing(file: AskFileResponse) -> None:
         # E.g., "abcd.csv" to "abcd_results.csv"
         result_file_name = file.name.removesuffix(".csv") + "_results.csv"
 
+        elements = [cl.File(name=result_file_name, path=result_file_path)]
         await cl.Message(
+            author="backend",
```
fg-nava, line 265: This is the main fix for the UI bug. Just needed the value for "author" key

app/src/chainlit.py
```diff
@@ -260,14 +260,25 @@ async def _batch_proccessing(file: AskFileResponse) -> None:
...
             content="File processed, results attached.",
-            elements=[cl.File(name=result_file_name, path=result_file_path)],
```
yoomlam, line 265: This can be restored, right?

app/src/chainlit.py
```diff
@@ -260,14 +260,25 @@ async def _batch_proccessing(file: AskFileResponse) -> None:
...
         ).send()
 
     except ValueError as err:
         await cl.Message(
             author="backend",
-            metadata={"error_class": err.__class__.__name__, "error": str(err)},
-            content=f"{err.__class__.__name__}: {err}",
+            content=f"Error processing file: {err}",
+            metadata={
+                "status": "processing_error",
```
yoomlam, line 280: `processing_error` sounds too generic to be useful. How about:
```suggestion
                "status": "batch_processing_error",
```

yoomlam (approved):

app/src/chainlit.py
fg-nava: Resolving as I removed the metadata key from Message objects

app/src/chainlit.py
```diff
@@ -248,26 +248,20 @@ def _get_retrieval_metadata(result: OnMessageResult) -> dict:
...
             content="File processed, results attached.",
             elements=[cl.File(name=result_file_name, path=result_file_path)],
         ).send()
 
     except ValueError as err:
         await cl.Message(
             author="backend",
-            metadata={"error_class": err.__class__.__name__, "error": str(err)},
-            content=f"{err.__class__.__name__}: {err}",
+            content=f"Error processing file: {err}",
```
ccheng26, line 266: not a blocker, but it would be helpful if we also added a logger for the error

ccheng26 (approved):

app/src/chainlit.py
fg-nava: Adding logging to `chainlit.py` and `batch_process.py`

PR #179, opened 2025-01-15
fix: Attempt to fix batch processing

yoomlam:
## Changes

Disable logging to LiteralAI while batch processing
Disable calling `msg.update()` during batch processing


yoomlam:
Fix applied to #181

PR #187, opened 2025-01-21
fix: Cleanup BOM characters from fieldnames in batch processing

fg-nava:
## Ticket
https://navalabs.atlassian.net/browse/DST-725

## Changes
- Modified CSV file opening to use `utf-8-sig` encoding
- Added cleanup step to strip BOM characters from column names
- Updated error logging for better debugging

## Context for reviewers
Excel-exported CSV files contain BOM characters that were causing the batch processor to fail. The fix ensures proper handling of these files by using UTF-8 with BOM signature encoding and cleaning up column names.

## Testing
Verified fix by:
1. Processing sample CSV provided by user with "question" column
2. Confirmed logs show clean column names without BOM characters
3. Batch processing completes successfully

### DIFF
--- a/app/src/batch_process.py
+++ b/app/src/batch_process.py
@@ -11,8 +11,15 @@ logger = logging.getLogger(__name__)
 
 
 async def batch_process(file_path: str, engine: ChatEngineInterface) -> str:
-    with open(file_path, mode="r", newline="", encoding="utf-8") as csvfile:
+    logger.info("Opening file: %s", file_path)
+
+    # Using utf-8-sig to handle CSV files with BOM characters from Excel
+    with open(file_path, mode="r", newline="", encoding="utf-8-sig") as csvfile:
         reader = csv.DictReader(csvfile)
+
+        # Clean up fieldnames to remove any BOM characters
+        if reader.fieldnames:
+            reader.fieldnames = [f.strip("\ufeff") for f in reader.fieldnames]
 
         if not reader.fieldnames or "question" not in reader.fieldnames:
             raise ValueError("CSV file must contain a 'question' column.")
--- a/app/src/batch_process.py
+++ b/app/src/batch_process.py
@@ -11,7 +11,6 @@ logger = logging.getLogger(__name__)
 
 
 async def batch_process(file_path: str, engine: ChatEngineInterface) -> str:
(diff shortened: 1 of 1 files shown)


yoomlam:
Does using `utf-8-sig` encoding work if the file is `utf-8` encoded? In other words, is using the `utf-8-sig` encoding backwards compatible to files without the BOM character?

fg-nava:
> Does using `utf-8-sig` encoding work if the file is `utf-8` encoded? In other words, is using the `utf-8-sig` encoding backwards compatible to files without the BOM character?

Yes, checked this:

https://stackoverflow.com/questions/63508421/is-utf-8-sig-suitable-for-decoding-both-utf-8-and-utf-8-bom

app/src/batch_process.py
```diff
@@ -11,9 +11,15 @@
...
 async def batch_process(file_path: str, engine: ChatEngineInterface) -> str:
-    with open(file_path, mode="r", newline="", encoding="utf-8") as csvfile:
+
+    # Using utf-8-sig to handle CSV files with BOM characters from Excel
+    with open(file_path, mode="r", newline="", encoding="utf-8-sig") as csvfile:
         reader = csv.DictReader(csvfile)
 
+        # Clean up fieldnames to remove any BOM characters
+        if reader.fieldnames:
+            reader.fieldnames = [f.strip("\ufeff") for f in reader.fieldnames]
```
yoomlam, line 21: This line bothers me since it's specifically for Excel-exported files. How about this approach instead: Re-encode file contents into `utf-8` before reading it with `DictReader` -- see https://stackoverflow.com/a/8898439/23458508. This makes it more obvious and compartmentalized, and you can add a code comment for the re-encoding code block that it's for Excel.

yoomlam:
An alternative is to ask the user to export non-BOM files -- https://superuser.com/a/1204249

app/src/batch_process.py
yoomlam: Yes! Good idea!

yoomlam (approved):

app/tests/src/util/test_file_util.py
yoomlam, line 1: Great! I didn't realize we didn't have unit tests for any of the `file_util` functions!

PR #216, opened 2025-02-18
feat: Move '(citation-N)' after punctuation

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-731

## Changes

* Add `move_citations_after_punctuation()` and unit test
* Updated Chatbot API, Batch Processing, and Chainlit UI to used new method
* Removed unused old method `reify_citations()`

## Testing

See unit test

Paragraphs:
image

Bullets:
image

### DIFF
--- a/app/src/chat_engine.py
+++ b/app/src/chat_engine.py
@@ -373,15 +373,15 @@ For example "List housing programs available to youth" or "What are the requirem
 Provide citation numbers:
 - When referencing the context, do not quote directly. Use the provided citation numbers (e.g., (citation-1)) to indicate when \
 you are drawing from the context. To cite multiple sources at once, you can append citations like so: (citation-1) (citation-2), etc. \
-For example: 'This is a sentence that draws on information from the context.(citation-1)'
+For example: 'This is a sentence that draws on information from the context. (citation-1)'
 
 Example question:
 Can a client get Unemployment and disability at the same time?
 
 Example Answer:
-No, a client cannot receive Unemployment Insurance (UI) and State Disability Insurance (SDI) benefits at the same time. (citation-1). \
+No, a client cannot receive Unemployment Insurance (UI) and State Disability Insurance (SDI) benefits at the same time. (citation-1)
 They must choose the program that best fits their situation. If they don't know which program to apply for, \
-they can apply for both, and their eligibility for each will be reviewed (citation-2) (citation-3)."""
+they can apply for both, and their eligibility for each will be reviewed. (citation-2) (citation-3)"""
 
     def on_message(
         self, question: str, chat_history: Optional[ChatHistory] = None
--- a/app/src/citations.py
+++ b/app/src/citations.py
@@ -216,6 +216,7 @@ def move_citations_after_punctuation(response: str) -> str:
     def move_citation(match: Match) -> str:
         citations = match.group(1)
         # match.group(2) only has the last citation in match.group(1)
+        # see https://stackoverflow.com/a/43866169/23458508
         punctuation = match.group(3)
         return f"{punctuation} {citations}\n"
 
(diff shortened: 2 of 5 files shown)


app/src/chat_engine.py
```diff
@@ -373,15 +373,15 @@ class ImagineLaEngine(BaseEngine):
...
 
 Example question:
 Can a client get Unemployment and disability at the same time?
 
 Example Answer:
-No, a client cannot receive Unemployment Insurance (UI) and State Disability Insurance (SDI) benefits at the same time. (citation-1). \
+No, a client cannot receive Unemployment Insurance (UI) and State Disability Insurance (SDI) benefits at the same time. (citation-1)
 They must choose the program that best fits their situation. If they don't know which program to apply for, \
-they can apply for both, and their eligibility for each will be reviewed (citation-2) (citation-3)."""
+they can apply for both, and their eligibility for each will be reviewed. (citation-2) (citation-3)"""
```
yoomlam, line 384: Updated citation format to what is desired.

fg-nava (approved):

yoomlam:
Not yet ready. I didn't handle multiple citations per sentence yet.

yoomlam:
Tried to match the `citation-N` pattern 1 or more times but it's only capturing the last iteration of the pattern. 
It appears I have to jump through more hoops: https://stackoverflow.com/a/43866169/23458508

app/tests/src/test_citations.py
```diff
@@ -183,3 +186,29 @@ def test_intro_sentence_for_list():
     assert subsections[1].text.startswith("The immigration officer will consider")
     assert "* Health" in subsections[1].text
     assert subsections[2].text.startswith("The officer weighs all these factors.")
+
+
+def test_move_citations_after_punctuation():
+    text = dedent(
```
KevinJBoyer, line 192: TIL about `dedent`!

app/src/citations.py
```diff
@@ -208,6 +208,21 @@ def replace_citation(match: Match) -> str:
     return re.sub(CITATION_PATTERN, replace_citation, response)
 
 
+def move_citations_after_punctuation(response: str) -> str:
+    """
+    After the '(citation-N)' should be a newline to avoid associating the citation with the next sentence.
+    """
```
KevinJBoyer, line 214: I'm not sure I'm following this comment -- is it saying that after this function is run, it will add newlines after every group of citations? Is this to prevent unintended behavior of re.sub?

KevinJBoyer: If so I'm wondering if we can avoid reformatting the text with newlines by using an arbitrary token that we temporarily inject (`<END CITATION GROUP>`) and then remove before returning the final text in this function

app/src/citations.py
```diff
@@ -208,6 +208,21 @@ def replace_citation(match: Match) -> str:
...
+
+    def move_citation(match: Match) -> str:
+        citations = match.group(1)
+        # match.group(2) only has the last citation in match.group(1)
```
KevinJBoyer, line 218: nice job working around this! Might be worth linking directly to the Stack Overflow site in the comments here.

KevinJBoyer (approved):
LGTM though had a question about the added newline behavior and a suggestion for how we might potentially avoid it

app/src/citations.py
yoomlam: The newline is for UI only, not due to any technical reason. My thinking: if `(citation-N)` appears after a punctuation and there's a subsequent sentence, then someone might associate `(citation-N)` with that subsequent sentence.
I can update the comment or remove the newline.

KevinJBoyer: Oh gotcha. Let's ask design what they think, I don't have an opinion one way or another. @RyanHansz - the suggestion is to add a newline after any citation so that the user sees citations like: 

A sentence. (1) (2)
Another sentence. (3)

instead of 

A sentence. (1) (2) Another sentence. (3)

RyanHansz: Same line seems to be the typical pattern, so `A sentence. (1) (2) Another sentence. (3)`

PR #218, opened 2025-02-19
fix: Support multiple datasets and k values using nargs in evaluation CLI

fg-nava:
## Ticket

N/A

## Changes

- Fixed `--dataset` argument in `app/src/metrics/cli.py` to handle multiple datasets and k values using `nargs="+"` in the make command for `run-evaluation`:
  - Also changed dataset filter parsing to split comma-separated values into a list
  - This (still) allows running evals on multiple datasets and/or k simultaneously (e.g., `...dataset="imagine_la la_policy", k="5 10"...`)

## Context for reviewers

When trying to evaluate multiple datasets, the command would fail because it was treating the comma-separated list as a single dataset name.

Fix modifies dataset arguments parsing, similar to how we handle k-values. We can pass multiple datasets using a comma-separated list.

Note: We currently only support 'imagine_la' and 'la_policy' datasets. This will be expanded to include other datasets (ca_wic, edd, etc.) as we add more evaluation data.

### DIFF
--- a/app/src/metrics/cli.py
+++ b/app/src/metrics/cli.py
@@ -4,16 +4,11 @@
 import argparse
 import json
 import os
-from typing import Any, Callable, List, Optional, Sequence
+from typing import Any, Callable, Optional, Sequence
 
 from src.retrieve import retrieve_with_scores
 
 from .evaluation.runner import run_evaluation
-
-
-def parse_k_values(k_str: str) -> List[int]:
-    """Parse comma-separated k values."""
-    return [int(k) for k in k_str.split(",")]
 
 
 def create_retrieval_function(min_score: float) -> Callable[[str, int], Sequence[Any]]:
--- a/app/src/metrics/cli.py
+++ b/app/src/metrics/cli.py
@@ -38,7 +38,7 @@ def main() -> None:
     parser.add_argument(
         "--dataset",
         type=str,
-        help="Dataset to evaluate (imagine_la, la_policy, or all)",
+        help="Comma-separated list of datasets to evaluate (e.g., imagine_la,la_policy or all)",
         required=True,
     )
     parser.add_argument("--k", type=str, default="5,10,25", help="Comma-separated list of k values")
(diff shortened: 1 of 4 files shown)


KevinJBoyer:
Instead of parsing this ourselves, could we use the nargs="+" functionality to allow a list to be passed in?

fg-nava:
> Instead of parsing this ourselves, could we use the nargs="+" functionality to allow a list to be passed in?

@KevinJBoyer is your suggestion just to convert the input types to nargs or is your suggestion that we should be agnostic to dataset name?

```
   # Current approach in batch.py:
   dataset_mapping = {
       "imagine_la": "Imagine LA",
       "la_policy": "LA County Policy",
   }
   ```
   
  [LOC](https://github.com/navapbc/labs-decision-support-tool/blob/c1594e1306171e3be60965118b3aa180e869172b/app/src/metrics/evaluation/batch.py#L129-L132)

**Comment by KevinJBoyer** [comment]
The former, and great catch re: also making the k parameter behave this way as well

File: app/src/metrics/cli.py [hunk 1]
```diff
@@ -83,7 +91,7 @@ def main() -> None:
     try:
         run_evaluation(
             questions_file=args.questions_file,
-            k_values=parse_k_values(args.k),
```
KevinJBoyer, line 84: We can clean up this function now too

app/tests/src/metrics/test_cli.py
```diff
@@ -64,16 +64,17 @@ def test_create_retrieval_function():
...
 )
 def test_main_dataset_filter(args, expected_dataset_filter):
     """Test main function handles dataset filter correctly."""
     with patch("argparse.ArgumentParser.parse_args") as mock_args:
         # Setup mock arguments
         mock_args.return_value = MagicMock(
-            dataset=args[1],
-            k="5,10",
+            dataset=args[1:] if len(args) > 1 else None,  # If no --dataset arg, return None
+            k=[5, 10],  # Now k is a list of integers
```
KevinJBoyer, line 77: nit: this comment makes sense in the context of this specific PR but won't to a future reader, suggest removing it and the equivalent on line 109

app/tests/src/metrics/test_cli.py
```diff
@@ -147,8 +147,8 @@ def test_main_results_display():
 
     with patch("argparse.ArgumentParser.parse_args") as mock_args:
         mock_args.return_value = MagicMock(
-            dataset="all",
-            k="5",
+            dataset=["all"],
```
KevinJBoyer, line 150: Does this need to be updated to be an empty list?

KevinJBoyer (approved):
LGTM w/ some remaining clean up, with caveat that I haven't tested locally!

PR #224, opened 2025-02-24
refactor: Consolidate evaluation metrics directory

fg-nava:
## Ticket
NA

## Changes
- Consolidated evaluation metrics directory structure by flattening nested directories
- Moved test files from nested directories to their new locations:
  - From `evaluation/metrics/evaluation/` to `evaluation/metrics/`
  - From `evaluation/metrics/models/` to `evaluation/metrics/`
  - From `evaluation/metrics/utils/` to `evaluation/utils/`
- Updated import statements across test files
- Removed old empty directories
- Updated CLI and runner code to match patterns from target branch

## Context for reviewers
Setting up evaluation metrics directory structure by removing unnecessary nesting.

## Testing
- make test


app/src/evaluation/metrics/runner.py
```diff
@@ -2,14 +2,38 @@
...
+    """
+
+    def retrieval_func(query: str, k: int) -> Sequence[Any]:
+        # Default to -1.0 if no min_score provided
+        score_threshold = min_score if min_score is not None else -1.0
+        return retrieve_with_scores(
+            query=query, retrieval_k=k, retrieval_k_min_score=score_threshold
+        )
+
+    return retrieval_func
```
KevinJBoyer, line 34: Consider using `functools.partial here`:
```suggestion
    from functools import partial
    score_threshold = min_score if min_score is not None else -1.0
    return partial(retrieve_with_scores, retrieval_k_min_score=score_threshold)
```

KevinJBoyer (approved):
LGTM with one suggestion around using functools.partial

app/src/evaluation/metrics/runner.py
fg-nava: Didn't end up including this suggestion since `results.batch_process_results` uses it as `func(query, k)`. But I agree your solution is more concise. I will look into applying this at a later time.

KevinJBoyer: > `results.batch_process_results` uses it as `func(query, k)`

That should still work, no?

PR #225, opened 2025-02-24
refactor: QA generation core functionality and evaluation system

fg-nava:
## Ticket

NA

## Changes

Refactors the evaluation system and adds a new QA generation module:

- Refactoring:
  - Consolidated evaluation metrics directory structure
  - Moved from metrics/models.py to data_models.py to include qa_generation models
  - Added content and dataset names to results.py and tests

- (New):
  - Added qa_generation module:
    - Generates QA pairs from documents using gpt-4o-mini 
    - Runner for executing QA generation jobs
    - Test files csovering `qa_generation`
  - qa_generation in CLI:
    - Added generate command
    - Updated CLI and Makefile to support generate-qa and evaluate

## Context for reviewers

Part of effort to automate QA pair generation and evaluation

## Testing

- make test
- make generate-qa dataset="imagine_la" sampling=0.05 && make evaluate dataset="imagine_la"


app/src/evaluation/qa_generation/config.py
```diff
@@ -0,0 +1,60 @@
...
+        """Get the directory for storing QA pairs with versioning."""
+        return self.output_dir / "qa_pairs" / self.version_id
+
+    @property
+    def latest_symlink(self) -> Path:
+        """Get the path to the 'latest' symlink."""
+        return self.output_dir / "qa_pairs" / "latest"
+
+    @classmethod
+    def from_cli_args(cls, args: Namespace) -> "GenerationConfig":
```
KevinJBoyer, line 52: Could we make this the constructor directly? I only see the current `__init__` being used in test

app/src/evaluation/cli/main.py
```diff
@@ -18,91 +19,149 @@
...
+        "--random-seed", type=int, help="Random seed for reproducible sampling"
+    )
+    eval_parser.add_argument(
+        "--commit", type=str, help="Git commit hash for tracking evaluation runs"
     )
 
     return parser
 
 
 def main() -> None:
```
KevinJBoyer, line 98: Threading somewhat arbitrarily, it seems like the code here could be simplified by splitting this up into two different files, `generate.py` and `evaluate.py`, and having the commands in the `Makefile` refer to these each separately, rather than both calling `main.py` and then branching on `if args.command == ...`.

app/src/evaluation/qa_generation/runner.py
```diff
@@ -0,0 +1,119 @@
...
+        sample_fraction: Fraction of documents to sample
+        random_seed: Random seed for reproducible sampling
+        git_commit: Git commit hash for tracking generation runs (not used in this simplified version)
+
+    Returns:
+        Path to generated QA pairs CSV
+    """
+    # Validate LLM model is specified
+    if not config.llm_model:
+        raise ValueError("No LLM model specified for QA generation")
```
KevinJBoyer, line 84: Is this possible?

KevinJBoyer (commented):
I haven't fully reviewed but a couple specific refactors that I wanted to share so you can start working on them while I continue reading:
 - `main.py` should be split into two different files, each with their own `main()`, since these are conceptually two different functionalities that each take their own set of arguments.
  - I don't think you need the QAGenerator and by extension GenerationConfig classes. QAGenerator is really a single function `generate_from_documents`, and the signature of this function is `generate_from_documents(llm_model: str, question_source: QuestionSource, questions_per_unit: int, documents: List[Document]) -> Iterator[QAPair]`

app/src/evaluation/qa_generation/runner.py
```diff
@@ -0,0 +1,119 @@
...
+                    "dataset",
+                    "created_at",
+                    "llm_model",
+                ],
+            )
+            writer.writeheader()
+            for pair in qa_pairs:
+                # Convert dataclass to dict
+                row = pair.__dict__.copy()
+                writer.writerow(row)
```
KevinJBoyer, line 56: You can use `writerows` to avoid the for loop, and [asdict](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) rather than `.__dict__.copy()`

app/src/evaluation/qa_generation/runner.py
```diff
@@ -0,0 +1,119 @@
...
+        if sample_fraction:
+            documents = get_stratified_sample(
+                documents,
+                sample_fraction=sample_fraction,
+                random_seed=random_seed,
+                key_func=lambda d: d.dataset,
+            )
+
+        # Generate QA pairs
+        qa_pairs = list(generator.generate_from_documents(documents))
```
KevinJBoyer, line 111: If we're going to collect all the results from the iterator in a list, we should have `generate_from_documents` just return a list

app/src/util/sampling.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Sample from each group
+    sampled_items = []
+    for group in groups.values():
+        sample_size = max(1, int(len(group) * sample_fraction))
+        sampled_items.extend(random.sample(group, sample_size))
+
+    # Reset random seed
+    if random_seed is not None:
+        random.seed()
```
KevinJBoyer, line 47: For my own understanding, why is this necessary?

app/src/util/sampling.py
```diff
@@ -0,0 +1,49 @@
...
+    if random_seed is not None:
+        random.seed(random_seed)
+
+    # Group items by key
+    groups: dict[str, List[T]] = {}
+    for item in items:
+        key = key_func(item)
+        if key not in groups:
+            groups[key] = []
+        groups[key].append(item)
```
KevinJBoyer, line 37: Can use collections.defaultdict to simply this:
```suggestion
    groups: dict[str, List[T]] = defaultdict(list)
    for item in items:
        key = key_func(item)
        groups[key].append(item)
```

app/tests/src/evaluation/qa_generation/test_generator.py
```diff
@@ -0,0 +1,271 @@
...
+from src.db.models.document import Chunk, Document
+from src.evaluation.data_models import QAPair
+from src.evaluation.qa_generation.config import GenerationConfig, QuestionSource
+from src.evaluation.qa_generation.generator import QAGenerator, generate_qa_pairs
+
+
+@pytest.fixture
+def mock_document():
+    """Create a mock document for testing."""
+    doc = MagicMock(spec=Document)
```
KevinJBoyer, line 20: Use `DocumentFactory` instead (which will give you lots of these default values for free)

app/tests/src/evaluation/qa_generation/test_generator.py
```diff
@@ -0,0 +1,271 @@
...
+def mock_completion_response():
+    """Create a mock completion response."""
+    response = MagicMock(spec=ModelResponse)
+    message = MagicMock()
+    message.content = json.dumps(
+        {
+            "question": "What is the test content?",
+            "answer": "The test content is for QA generation.",
+        }
+    )
```
KevinJBoyer, line 53: Can this behavior happen outside of test? (Related to my earlier comment about `num_pairs`)

app/tests/src/evaluation/qa_generation/test_generator.py
```diff
@@ -0,0 +1,271 @@
...
+def test_generate_qa_pairs_list_format(mock_document):
+    """Test handling of list format in LLM response."""
+    response = MagicMock(spec=ModelResponse)
+    message = MagicMock()
+    message.content = json.dumps(
+        [
+            {"question": "Question 1?", "answer": "Answer 1"},
+            {"question": "Question 2?", "answer": "Answer 2"},
+        ]
+    )
```
KevinJBoyer, line 105: Can this behavior happen outside of test? (Related to my earlier comment about num_pairs)

app/tests/src/evaluation/qa_generation/test_generator.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def test_qa_generator_get_chunks_document_level():
+    """Test getting chunks at document level."""
+    config = GenerationConfig(question_source=QuestionSource.DOCUMENT, questions_per_unit=2)
+    generator = QAGenerator(config)
+
+    doc1 = MagicMock(spec=Document)
+    doc2 = MagicMock(spec=Document)
+    documents = [doc1, doc2]
```
KevinJBoyer, line 208:
```suggestion
    documents = DocumentFactory.build_batch(2)
```

app/tests/src/evaluation/qa_generation/test_generator.py
```diff
@@ -0,0 +1,271 @@
...
+def test_qa_generator_init():
+    """Test QA generator initialization."""
+    config = GenerationConfig(question_source=QuestionSource.DOCUMENT, questions_per_unit=2)
+    generator = QAGenerator(config)
+
+    assert generator.config == config
+    assert generator.llm == config.llm_model
+
+
+def test_qa_generator_get_chunks_document_level():
```
KevinJBoyer, line 201: On testing strategy: `_get_chunks_to_process` is a private method used by `generate_from_documents`. Rather than cover `_get_chunks_to_process` with a unit test, it's generally better to write a test that covers `generate_from_documents` in such a way that whatever behavior `generate_from_documents` depends on `_get_chunks_to_process` for is covered by the test on `generate_from_documents`. It's less code to write up front, but more importantly, `_get_chunks_to_process` is an implementation detail rather than an invariant we want to preserve: if a future maintainer wants to rewrite `generate_from_documents` without using it, that shouldn't break any tests.

app/tests/src/evaluation/qa_generation/test_generator.py
```diff
@@ -0,0 +1,271 @@
...
+
+def test_qa_generator_generate_from_documents(mock_document, mock_completion_response):
+    """Test generating QA pairs from documents."""
+    config = GenerationConfig(question_source=QuestionSource.DOCUMENT, questions_per_unit=1)
+    generator = QAGenerator(config)
+
+    with patch(
+        "src.evaluation.qa_generation.generator.generate_qa_pairs",
+        return_value=[MagicMock(spec=QAPair)],
+    ):
```
KevinJBoyer, line 254: Why do we need this mock?

KevinJBoyer (commented):
Finished reading through -- the logic overall looks good, most of these comments are around code structure and hygiene. Nice work on this!

app/src/evaluation/cli/main.py
fg-nava: This is a helpful recommendation, thanks. I have this completed in a local commit and will push soon.

Do you think we should split up `data_models.py` between "generate" and "evaluate" steps? My initial thought with `main.py` was to write merge logic at the level of evaluation for CLI and type definitions, but I see if the benefit of keeping steps isolated and calling them as make targets.

app/src/evaluation/utils/id_generator.py
```diff
@@ -0,0 +1,9 @@
+import uuid
+from hashlib import md5
+
+
+def generate_stable_id(question: str, answer: str) -> uuid.UUID:
+    """Generate a stable UUID for a QA pair based on content."""
+    content = f"{question}||{answer}".encode("utf-8")
+    content_hash = md5(content, usedforsecurity=False).digest()
+    return uuid.UUID(bytes=content_hash[:16])
```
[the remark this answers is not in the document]
fg-nava: I do think MD5 is more transparent to what the ID represents. I had a preference to maintain the UUID standard in the case of using this as a PK for future database inserts/updates - open to switching if preferred!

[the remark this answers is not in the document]
KevinJBoyer: ah that makes sense, thanks for explaining!

app/src/evaluation/cli/main.py
KevinJBoyer: I think data_models.py is fine either way -- the idea here is to try to minimize the amount of nesting that someone has to keep in their head while following the code

app/src/evaluation/qa_generation/generator.py
```diff
@@ -0,0 +1,188 @@
...
+}
+Do not include any additional formatting, newlines, or text outside the JSON.
+"""
+
+MAX_WORKERS = 5  # Limit concurrent API calls
+
+
+def generate_qa_pairs(
+    document_or_chunk: Document | Chunk, num_pairs: int = 1, llm: str = "gpt-4o-mini"
+) -> List[QAPair]:
```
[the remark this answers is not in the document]
fg-nava: The initial process in the QA Generation notebook was set up for the option of multiple pair per chunk: 

```
num_qa_per_chunk_or_doc = widgets.IntText(
    value=5,
    description='Number of questions per chunk/document:',
    disabled=False   
)
```

My intent was to keep this as an argument if we wanted to test multiple pairs per chunk. That said, I've since made it so the arg is not exposed to main functions. I'm done with a change (still local) that removes num_pairs and questions_per_unit.

app/src/evaluation/qa_generation/generator.py
```diff
@@ -0,0 +1,188 @@
...
+                        generated_pairs.append(pair)
+                    except json.JSONDecodeError:
+                        continue
+
+        # Validate required fields
+        valid_pairs = []
+        for pair in generated_pairs:
+            if "question" in pair and "answer" in pair:
+                valid_pairs.append(pair)
+        generated_pairs = valid_pairs
```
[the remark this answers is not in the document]
fg-nava: Using structured outputs now

app/src/evaluation/data_models.py
```diff
@@ -95,3 +98,21 @@ class MetricsSummary:
...
     dataset_metrics: Dict[str, DatasetMetrics]
     incorrect_analysis: IncorrectRetrievalsAnalysis
+
+
+# QA Generation Models
+@dataclass
+class QAPair:
+    """A question-answer pair."""
+
+    id: UUID  # Stable ID generated at creation time
```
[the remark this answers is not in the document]
fg-nava: Using in data_models.__post_init__ now

app/tests/src/evaluation/qa_generation/test_generator.py
fg-nava: Using `tests.src.db.models.factories` now

app/tests/src/evaluation/qa_generation/test_generator.py
fg-nava: Removed tests for edgecases when multiple QA pairs can be generated from single LLM response

app/src/util/sampling.py
fg-nava: The reset makes sure that a seed() used somewhere else has fresh randomness

https://stackoverflow.com/questions/11929701/why-is-seeding-the-random-generator-not-stable-between-versions-of-python

fg-nava:
> I don't think you need the QAGenerator and by extension GenerationConfig classes. QAGenerator is really a single function generate_from_documents, and the signature of this function is generate_from_documents(llm_model: str, question_source: QuestionSource, questions_per_unit: int, documents: List[Document]) -> Iterator[QAPair]

Addressed in [commit](https://github.com/navapbc/labs-decision-support-tool/pull/225/commits/3dfe77acd91898f8d5c28e5faffa5b52e03b6f07)

app/tests/src/evaluation/qa_generation/test_generator.py
fg-nava: New tests now cover generate_from_documents

app/tests/src/evaluation/qa_generation/test_qa_generation_runner.py
```diff
@@ -0,0 +1,281 @@
+"""Tests for QA generation runner."""
```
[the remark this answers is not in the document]
fg-nava: Using tmp_path but new changes test actual output files.

KevinJBoyer (approved):
Left you a note in Slack about the test coverage in test_generator and test_qa_generation_runner, but all good to merge from my end!

app/src/evaluation/qa_generation/runner.py
fg-nava: LOC removed now

app/tests/src/evaluation/qa_generation/test_generator.py
fg-nava: Fixed in latest commit

app/src/evaluation/qa_generation/generator.py
```diff
@@ -0,0 +1,135 @@
...
+                    "content": GENERATE_QUESTION_ANSWER_PROMPT,
+                    "role": "system",
+                },
+                {
+                    "content": f"Please create one high-quality question-answer pair from this content and format it as JSON: {document_or_chunk.content}",
+                    "role": "user",
+                },
+            ],
+            temperature=app_config.temperature,
+            response_format={"type": "json_object"},
```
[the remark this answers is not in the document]
fg-nava: Added

fg-nava:
Thanks for the approval -- I'm going to keep this PR open until I fix overused mocks in other testfiles in /cli and /metrics

fg-nava:
> Thanks for the approval -- I'm going to keep this PR open until I fix overused mocks in other testfiles in /cli and /metrics

As an update, merging as is. Current status: `qa_generation` tests are in good shape. 
To do: fix `/cli` and `/metrics/` tests in next PR

PR #229, opened 2025-02-26
fix: Enable multiple user messages per LiteralAI thread

yoomlam:
## Ticket

## Changes

Chat history for a single session is now correctly reflected in LiteralAI.
https://nava.slack.com/archives/C06DP498D1D/p1740507232929499

## Testing

```
curl -X POST 'http://0.0.0.0:8000/api/query' -H 'Content-Type: application/json' -d '{ "session_id": "A123", "new_session": true, "message": "what do you know about?"}'
curl -X POST 'http://0.0.0.0:8000/api/query' -H 'Content-Type: application/json' -d '{ "session_id": "A123", "new_session": false, "message": "list programs"}'
curl -X POST 'http://0.0.0.0:8000/api/query' -H 'Content-Type: application/json' -d '{ "session_id": "A123", "new_session": false, "message": "Thanks"}'
```

Check in LiteralAI
image

### DIFF
--- a/app/src/chat_api.py
+++ b/app/src/chat_api.py
@@ -320,14 +320,14 @@ async def query(request: QueryRequest) -> QueryResponse:
     return response
 
 
-def _validate_session_against_literalai(request, session):
+def _validate_session_against_literalai(request: QueryRequest, session: UserSession) -> None:
     thread_id = session.literalai_thread_id
     # Check if request is consisten with LiteralAI thread
     if request.new_session:
         if thread_id:
             raise HTTPException(
                 status_code=409,
-                detail=f"Cannot start a new session with an existing session_id: {request.session_id} associated with thread_id: {thread_id}",
+                detail=f"Cannot start a new session '{request.session_id}' that is already associated with thread_id: {thread_id}",
             )
     else:
         if not thread_id:
@@ -337,7 +337,7 @@ def _validate_session_against_literalai(request, session):
             )
 
 
-def _validate_literalai_message(session, request_msg):
+def _validate_literalai_message(session: UserSession, request_msg: ChatMessage) -> None:
     if not session.literalai_thread_id:
         if not request_msg.thread_id:
             raise HTTPException(
--- a/app/src/chat_api.py
+++ b/app/src/chat_api.py
@@ -66,29 +66,46 @@ class ChatEngineSettings:
 @dataclass
 class UserSession:
     user: UserInfo
+    session_id: str
     chat_engine_settings: ChatEngineSettings
     literalai_user_id: Optional[str] = None
+    literalai_thread_id: Optional[str] = None
     message_history: Optional[Sequence[ChatMessage]] = None
 
 
-def __query_user_session(user_id: str) -> UserSession:
+# FIXME: Replace this with a real DB table
+DB_SESSIONS: dict[str, UserSession] = {}
+
+
+def __query_user_session(user_id: str, session_id: str) -> UserSession:
     """
-    Placeholder for creating/retrieving user's session from the DB, including settings and constraints
+    FIXME: Placeholder for creating/retrieving user's session from the DB, including settings and constraints
     """
+    if session_id in DB_SESSIONS:
+        logger.info("Found user session for: %s", user_id)
+        return DB_SESSIONS[session_id]
+
     session = UserSession(
         user=UserInfo(user_id, ["imagine-la"]),
+        session_id=session_id,
         chat_engine_settings=ChatEngineSettings("imagine-la"),
     )
+    DB_SESSIONS[session_id] = session
     logger.info("Found user session for: %s", user_id)
     return session
 
 
-async def _get_user_session(user_id: str) -> UserSession:
-    session = __query_user_session(user_id)
+async def _get_user_session(user_id: str | None, session_id: str) -> UserSession:
+    user_id = user_id or "Anonymous"
+    session = __query_user_session(user_id, session_id)
     # Ensure user exists in Literal AI
     literalai_user = await literalai().api.get_or_create_user(user_id, session.user.__dict__)
     # Set the LiteralAI user ID for this session so it can be used in literalai().thread()
     session.literalai_user_id = literalai_user.id
+    print(f"User {user_id} has LiteralAI ID: {literalai_user.id}")
+    print(
+        f"Session {session.session_id} is associated with LiteralAI thread_id: {session.literalai_thread_id}"
+    )
     return session
 
 
@@ -108,7 +125,7 @@ def _load_chat_history(db_session: db.Session, session_id: str) -> ChatHistory:
 # Make sure to use async functions for faster responses
 @router.get("/engines")
 async def engines(user_id: str) -> list[str]:
-    session = await _get_user_session(user_id)
+    session = await _get_user_session(user_id, "")
     # Example of using Literal AI to log the request and response
     with literalai().thread(name="API:/engines", participant_id=session.literalai_user_id):
         request_msg = literalai().message(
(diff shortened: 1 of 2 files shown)


app/src/chat_api.py
```diff
@@ -66,29 +66,46 @@ class ChatEngineSettings:
...
 
-def __query_user_session(user_id: str) -> UserSession:
+# FIXME: Replace this with a real DB table
+DB_SESSIONS: dict[str, UserSession] = {}
+
+
+def __query_user_session(user_id: str, session_id: str) -> UserSession:
     """
-    Placeholder for creating/retrieving user's session from the DB, including settings and constraints
+    FIXME: Placeholder for creating/retrieving user's session from the DB, including settings and constraints
```
yoomlam, line 82: Addressing these `FIXME` will be in the next PR

app/src/chat_api.py
```diff
@@ -151,17 +168,8 @@ async def feedback(
         response_id: the response_id of the chatbot response
         comment: user comment for the feedback
         user_id: the user's id
-
-    Returns:
-        FeedbackResponse
-        user_id: the user's id
-        value: 1 if is_positive was "true" and 0 if is_positive was "false"
-        step_id: ID of the step associated with the score
-        comment: the initial user comment for the feedback
```
yoomlam, line 160: This comment is incorrect. Removing it to be consistent with the code.

app/src/chat_api.py
```diff
@@ -108,7 +125,7 @@ def _load_chat_history(db_session: db.Session, session_id: str) -> ChatHistory:
 # Make sure to use async functions for faster responses
 @router.get("/engines")
 async def engines(user_id: str) -> list[str]:
-    session = await _get_user_session(user_id)
+    session = await _get_user_session(user_id, "")
```
yoomlam, line 127: Session id is irrelevant for this query

app/src/chat_api.py
```diff
@@ -242,39 +250,52 @@ def get_chat_engine(session: UserSession) -> ChatEngineInterface:
...
+    one_line_question = None
+    if request.new_session:
+        # Only set the LiteralAI thread name if it's a new session; otherwise the thread name will be updated with the last message
+        one_line_question = request.message.strip().splitlines()[0] or "API:/query"
     with (
-        literalai().thread(name=one_line_question, participant_id=session.literalai_user_id),
+        literalai().thread(
+            name=one_line_question,
+            participant_id=session.literalai_user_id,
+            thread_id=session.literalai_thread_id,
```
yoomlam, line 263: Crux: need to pass a `thread_id` to the `thread()` context manager.

app/src/chat_api.py
```diff
@@ -242,39 +249,55 @@ def get_chat_engine(session: UserSession) -> ChatEngineInterface:
...
+                "request": request.__dict__,
+                "user": session.user.__dict__,
+                "chat_history": chat_history,
+            },
+        )
+
+        _validate_literalai_message(session, request_msg)
+        if not session.literalai_thread_id and request_msg.thread_id:
+            # session.literalai_thread_id is None when request.new_session=True
+            session.literalai_thread_id = request_msg.thread_id
```
yoomlam, line 299: Capture the `thread_id` for subsequent queries

app/src/chat_api.py
```diff
@@ -299,14 +322,47 @@ async def query(request: QueryRequest) -> QueryResponse:
...
+    if not request.new_session and not session.literalai_thread_id:
+        raise HTTPException(
+            status_code=409,
+            detail=f"LiteralAI thread ID for existing session {request.session_id!r} not found",
+        )
+
+
+def _validate_literalai_message(session: UserSession, request_msg: Message) -> None:
+    if not request_msg.thread_id:
+        # Should never happen
```
yoomlam, line 345: `thread_id` should be populated when we get `request_msg` from LiteralAI. If it's ever `None`, we should see an obvious exception.

app/tests/src/test_chat_api.py
```diff
@@ -73,11 +73,9 @@ async def create_score(
 
 
 @pytest.fixture
-def literalai_client(monkeypatch):
```
yoomlam, line 76: This returns a TestClient, not a LiteralAI client.

app/tests/src/test_chat_api.py
yoomlam, line 1: Cleaned up this test file

app/src/chat_api.py
```diff
@@ -66,29 +66,45 @@ class ChatEngineSettings:
...
+    literalai_thread_id: Optional[str] = None
     message_history: Optional[Sequence[ChatMessage]] = None
 
 
-def __query_user_session(user_id: str) -> UserSession:
+# FIXME: Replace this with a real DB table
+DB_SESSIONS: dict[str, UserSession] = {}
+
+
+def __query_user_session(user_id: str, session_id: str) -> UserSession:
```
KevinJBoyer, line 80: I know this is present in the original code, but why the double underscore?

app/src/chat_api.py
KevinJBoyer: Can we instead change the signature of `_get_user_session` to be `_get_user_session(user_id: str | None, session_id: str | None)` rather than passing a blank string? I'm also not sure about the behavior here, it seems like every call to /engines will create a new UserSession with a blank session ID -- is that desired?

app/src/chat_api.py
```diff
@@ -151,17 +167,8 @@ async def feedback(
...
-        FeedbackResponse
-        user_id: the user's id
-        value: 1 if is_positive was "true" and 0 if is_positive was "false"
-        step_id: ID of the step associated with the score
-        comment: the initial user comment for the feedback
     """
-    user_session_id = request.user_id if request.user_id else request.session_id
-
-    session = await _get_user_session(user_session_id)
+    session = await _get_user_session(request.user_id, request.session_id)
```
KevinJBoyer, line 171: For my own understanding, why is session needed here? Isn't `session.user.user_id` guaranteed to be the same value as `request.user_id`?

app/src/chat_api.py
KevinJBoyer: And the literalai_thread_id is generated by Literal AI (not under our control) by the `request_msg = literalai().message(` call on line 285 below, is that right?

app/src/chat_api.py
```diff
@@ -242,39 +249,55 @@ def get_chat_engine(session: UserSession) -> ChatEngineInterface:
...
-            content=request.message,
-            type="user_message",
-            name=request.session_id,
-            metadata={
-                "request": request.__dict__,
-                "user": session.user.__dict__,
-            },
-        )
-        # Load history BEFORE saving the new message
+        # Load history BEFORE adding the new message to the DB
```
KevinJBoyer, line 268: Do we still need to do this (i.e., include `"chat_history": chat_history` in the metadata) if we're correctly using threads? Ah, reading the code more carefully, is this so that we can properly raise an exception based on the `request.new_session` parameter?

app/src/chat_api.py
KevinJBoyer: thank you for including this!

KevinJBoyer (approved):
LGTM -- left a few questions just to make sure I understand everything right.

app/src/chat_api.py
yoomlam: Expresses that this is *really* not meant to be used outside this class, and for me to see this is only called by another underscore function.

app/src/chat_api.py
yoomlam: Sure, I'll change it.
> is that desired?
Yes, a blank session id is fine. There's no chat session associated with this API call.

app/src/chat_api.py
yoomlam: I think providing `request.user_id` was part of API spec that Charlie discussed with SBN.

app/src/chat_api.py
yoomlam: Yes

yoomlam: We can remove it from metadata if desired.

yoomlam:
When trying to use `session_id` as the `thread_id`, here's the error I get when using session_id `H123` (the API client determines the `session_id`):
```
  File "/home/yoom/.cache/pypoetry/virtualenvs/decision-support-tool-9TtSrW0h-py3.12/lib/python3.12/site-packages/literalai/api/__init__.py", line
182, in raise_error
    raise Exception(error)
Exception: Failed to send steps: {"errors":[{"message":"The field must be a valid id string of length 16 to 128 characters comprised of alphanumeric
or -","locations":[{"line":19,"column":7}],"path":["step0"],"extensions":{"code":"INVALID_ARGS","originalError":[{"code":"custom","message":"The
field must be a valid id string of length 16 to 128 characters comprised of alphanumeric or -","fatal":true,"path":["threadId"]}]}}],"data":null}
```
Since there are some constraints on the `thread_id`, we'll let LiteralAI assign the `thread_id` and just use what they generate as implemented in this PR.

PR #230, opened 2025-02-26
feat: Add UserSession DB table

yoomlam:
## Ticket

Follow-on to #229 

## Changes

* Add UserSession DB table
* Refactor to consider DB transactions

## Testing

Same as #229, plus check your local DB as you are running each command.


app/src/chat_api.py
```diff
@@ -51,70 +51,99 @@ def literalai() -> AsyncLiteralClient:
...
+    return chat_session
+
+
+def _load_user_session(session_id: str | None) -> Optional[UserSession]:
+    if not session_id:
+        return None
+    with app_config.db_session() as db_session:
+        return db_session.scalars(
+            select(UserSession).where(UserSession.session_id == session_id)
+        ).first()
```
KevinJBoyer, line 136: more an FYI than a suggestion, but you can use .one() or .one_or_none() to get an exception "for free" if there's unexpectedly more than one result returned (which I don't think can be the case here anyways because session_id is a primary key)

app/src/chat_api.py
```diff
@@ -124,55 +153,42 @@ def _load_chat_history(db_session: db.Session, session_id: str) -> ChatHistory:
...
-        user_id: the user's id
-    """
-    session = await _get_user_session(request.user_id, request.session_id)
+    """Endpoint for creating feedback for a chatbot response message"""
+    session = await _get_chat_session(request.user_id, request.session_id)
     # API endpoint to send feedback https://docs.literalai.com/guides/logs#add-a-score
     response = await literalai().api.create_score(
         step_id=request.response_id,
-        name=session.user.user_id,
+        name=request.user_id or session.user_session.user_id,
```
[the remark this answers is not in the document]
KevinJBoyer: With the other PR I believe request.user_id should always be set now

app/src/chat_api.py
```diff
@@ -233,71 +249,63 @@ class QueryResponse(BaseModel):
...
+    session = await _get_chat_session(request.user_id, request.session_id, user_meta)
     _validate_session_against_literalai(request, session)
 
-    one_line_question = None
+    # Load history BEFORE adding the new message to the DB
+    chat_history = _load_chat_history(request.session_id)
+    _validate_chat_history(request, chat_history)
+
+    thread_name = None
+    # Only if new session, set the LiteralAI thread name; don't want the thread name to change otherwise
```
KevinJBoyer, line 276: So to confirm, passing None to literalai().thread(name=...) leaves the existing thread name the same?

KevinJBoyer (approved):
LGTM! Thanks for being game to try the approach of using the API session_id as the thread ID even though it didn't work out!

app/src/db/models/conversation.py
```diff
@@ -15,3 +15,16 @@ class ChatMessage(Base, IdMixin, TimestampMixin):
     )
     role: Mapped[str] = mapped_column(comment="Role of the message speaker")
     content: Mapped[str] = mapped_column(comment="Content of the message")
+
+
+class UserSession(Base, TimestampMixin):
```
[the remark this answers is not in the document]
yoomlam: #232

app/src/chat_api.py
yoomlam: TIL

PR #232, opened 2025-02-26
feat: DB: Add session_id foreign key constraint

yoomlam:
## Ticket

Follow-on to #230 

## Changes

* Add `session_id` foreign key constraint between `UserSession` and `ChatMessage` tables
* Use `UserSession.chat_messages` to get associated messages

## Testing
Check the contents of the `user_session` and `chat_message` DB tables, particularly the `session_id` values.
After running:
```
make db-migrate
```
There should be new records in `user_session` so that all `session_id` values in the `chat_message` DB table have a corresponding `user_session` record.

Same as #229, plus check your local DB as you are running each command.


app/src/db/models/conversation.py
```diff
@@ -28,3 +30,7 @@ class UserSession(Base, TimestampMixin):
     lai_thread_id: Mapped[str | None] = mapped_column(
         comment="LiteralAI's thread ID corresponding to session_id"
     )
+    chat_messages: Mapped[list[ChatMessage]] = relationship(
+        backref="ChatMessage.session_id",
```
KevinJBoyer, line 34: To avoid backref (which is deprecated) I think you can do:
```suggestion
        back_populates="ChatMessage.session_id",
```
here and then add `session: Mapped["UserSession"] = relationship("UserSession", back_populates="chat_messages")` to ChatMessage above

KevinJBoyer (approved):
LGTM, thoughtful approach to migrating when we have existing records!

app/src/db/models/conversation.py
yoomlam: updated

PR #233, opened 2025-02-26
chore: API: return error on empty user_id

yoomlam:
## Ticket

https://nava.slack.com/archives/C06DP498D1D/p1740606102296019

## Changes

Return HTTP error upon empty `user_id`

## Testing

Added unit test

### DIFF
--- a/app/src/chat_api.py
+++ b/app/src/chat_api.py
@@ -8,10 +8,10 @@ import functools
 import logging
 import uuid
 from dataclasses import dataclass
-from typing import Optional, Sequence
+from typing import Annotated, Optional, Sequence
 
 from asyncer import asyncify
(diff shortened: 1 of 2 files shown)


app/src/chat_api.py
```diff
@@ -111,6 +111,8 @@ async def _get_chat_session(
     session_id: str | None,
     user_meta: Optional[dict] = None,
 ) -> ChatSession:
+    if not user_id:
+        raise HTTPException(status_code=400, detail="user_id must be a non-empty string")
```
yoomlam, line 115: Backup code in case this method is called anywhere else (in future PRs).

fg-nava (approved):

PR #235, opened 2025-02-27
fix: Temporarily revert "chore: API: return error on empty user_id"

KevinJBoyer:
Revert #233 to unblock testing while empty user id issue is worked on the front end

### DIFF
--- a/app/src/chat_api.py
+++ b/app/src/chat_api.py
@@ -8,10 +8,10 @@ import functools
 import logging
 import uuid
 from dataclasses import dataclass
-from typing import Annotated, Optional, Sequence
+from typing import Optional, Sequence
 
 from asyncer import asyncify
-from fastapi import APIRouter, HTTPException, Query, Request, Response
+from fastapi import APIRouter, HTTPException, Request, Response
 from literalai import AsyncLiteralClient, Message
 from pydantic import BaseModel
 from sqlalchemy import select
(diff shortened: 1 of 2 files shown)


fg-nava (approved):

fg-nava:
I checked out this branch and ran this curl:
`curl -X POST "http://localhost:8000/api/query" -H "Content-Type: application/json" -d '{"session_id": "test-session-123", "new_session": true, "message": "Hello", "user_id": ""}' -v`

Returns:
```
* Connected to localhost (::1) port 8000
> POST /api/query HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/8.7.1
> Accept: */*
> Content-Type: application/json
> Content-Length: 90
> 
} [90 bytes data]
* upload completely sent off: 90 bytes
< HTTP/1.1 500 Internal Server Error
< date: Thu, 27 Feb 2025 18:16:55 GMT
< server: uvicorn
< content-length: 21
< content-type: text/plain; charset=utf-8
```

#233 handles this error with a proper 400:

```
{
  "detail": [
    {
      "type": "missing",
      "loc": [
        "body",
        "user_id"
      ],
      "msg": "Field required",
      "input": {
        "session_id": "test-session-123",
        "new_session": true,
        "message": "Hello"
      }
    }
  ]
}
```

I think the fix here is to make a fix downstream of #233 with a temporarily permissive/fallback user-id. Wdyt?

KevinJBoyer:
Yup, you're right!

PR #241, opened 2025-03-03
feat: Facilitate 'all' datasets in refresh-ingestion.sh

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-809

## Changes

- Added `all` argument to `refresh-ingestion.sh` to facilitate ingestion for all datasets for both dev and prod
- Added `--drop-only` ingest option to only drop the dataset so that ingestion can use `--resume` in subsequent runs
- Save AWS-related commands to separate script files to be executed separately since they affect the deployments and some commands can fail due to AWS resources.

## Testing

```
./refresh-ingestion.sh all
```
then follow the instructions
```
- Review and run the refresh scripts (in the top-level folder): ...
- Restore local TF to dev environment: ./bin/terraform-init infra/app/service dev
```

### DIFF
--- a/app/refresh-ingestion.sh
+++ b/app/refresh-ingestion.sh
@@ -246,11 +246,6 @@ check_preconditions(){
         echo "Move or delete the file/folder(s) before running this script."
         exit 50
     fi
-
-    if ! aws sts get-caller-identity; then
-        echo "ERROR: AWS CLI is not configured. Run 'aws configure' to set up."
-        exit 51
-    fi
 }
 
 if ! [ -e "refresh-ingestion.sh" ]; then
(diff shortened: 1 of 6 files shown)


app/refresh-ingestion.sh
```diff
@@ -195,21 +244,41 @@ check_preconditions(){
...
 if [ -z "$1" ]; then
     echo "Usage: '$0 <DATASET_ID>', where <DATASET_ID> is any of the following:"
     echo -n "  imagine_la"
     grep "case" src/ingest_runner.py | tr '":' ' ' | while read CASE DATASET_ID COLON; do [ "$DATASET_ID" != "_" ] && echo -n ", $DATASET_ID"; done
     echo ""
+    echo "If <DATASET_ID> is 'all', it will run re-scrape all the datasets and create scripts for both dev and prod."
+    echo "  SKIP_LOCAL_EMBEDDING=true by default when <DATASET_ID>='all'."
+    echo "Set DEPLOY_ENV to 'dev' (default) or 'prod' to create refresh-$DEPLOY_ENV.sh script for the appropriate environment."
+    echo "Set SKIP_LOCAL_INGEST=true to skip local ingestion but still create files for upload and ingestion for the deployed app."
+    echo "Set TODAY=YYYY-MM-DD to use a different date than the current day."
```
fg-nava, line 264: JW, when would we set this variable to a date other than today?

fg-nava (approved):

PR #243, opened 2025-03-04
feat: Back up and restore DB contents

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-819

## Changes

* Update `main-app` image with updated Postgres version to be compatible with Postgres 16 in the `main-db` image so that `pg_dump` and `pg_restore` can connect to `main-db`
* Add `pg_util.py` to dump and restore the DB
* Updated `Makefile` and `pyproject.toml` with new targets/scripts

## Testing

```
make build
make start-db
make pg-dump args="backup --dumpfile db2.dump"

make pg-dump args="restore --dumpfile db2.dump"
```

After this PR merges, run `./bin/run-command app dev '["pg-dump", "backup"]'`, download from [S3](https://us-east-1.console.aws.amazon.com/s3/buckets/decision-support-tool-app-dev?region=us-east-1&bucketType=general&prefix=pg_dumps%2F&tab=objects), and restore.


app/src/db/pg_util.py
```diff
@@ -0,0 +1,202 @@
...
+
+def backup_db() -> None:
+    env = os.environ.get("ENVIRONMENT", "local")
+    dumpfilename = os.environ.get("PG_DUMP_FILE", f"{env}_db.dump")
+    if os.path.exists(dumpfilename):
+        logger.fatal(
+            "File %r already exists; please delete it first or set PG_DUMP_FILE to a nonexistent file",
+            dumpfilename,
+        )
+        return
```
KevinJBoyer, line 32: What's the reason to use an env var for configuration vs. accepting an optional argument? (`make backup-db output.dump`)? That'd let you remove the `DOCKER_EXEC_ARGS := -e PG_DUMP_FILE -e TRUNCATE_TABLES` from the Makefile, too

app/src/db/pg_util.py
```diff
@@ -0,0 +1,202 @@
...
+
+    config_dict = _get_db_config()
+    _print_row_counts()
+
+    # In local environments, write to the current directory
+    if env == "local":
```
KevinJBoyer, line 38: Could you use `smart_open` here to avoid needing to branch on the environment (besides setting the output path based on local vs. dev/prod)?

KevinJBoyer: e.g.:
```
dumpfilename = ... if env == "local" else ...
if not _pg_dump(config_dict, dumpfilename): # _pg_dump modified to use smart_open
    ...

File: app/src/db/pg_util.py [hunk 3]
```diff
@@ -0,0 +1,202 @@
...
+        )
+        dest_path = f"pg_dumps/{dated_filename}"
+        try:
+            s3_client.upload_file(stdout_file, bucket, dest_path)
+            logger.info("DB dump uploaded to s3://%s/%s", bucket, dest_path)
+        except ClientError as e:
+            logging.error(e)
+
+
+TRUE_STRINGS = ["true", "1", "t", "y", "yes"]
```
KevinJBoyer, line 71: Since this is under our control, let's set a strict requirement for exactly what we want to receive.

app/src/db/pg_util.py
```diff
@@ -0,0 +1,202 @@
...
+    if not os.path.exists(dumpfilename):
+        logger.fatal("File %r not found; please set PG_DUMP_FILE to a valid file", dumpfilename)
+        return
+
+    config_dict = _get_db_config()
+    _print_row_counts()
+
+    truncate_tables = os.environ.get("TRUNCATE_TABLES", "True").strip().lower() in TRUE_STRINGS
+    if truncate_tables:
+        delay = "TRUNCATE_TABLES" not in os.environ
```
KevinJBoyer, line 85: I think the intent here is to prevent someone from accidentally truncating with restore-db -- if that's the case, could we flip this so the user has to opt in to truncation: e.g., `make restore-db args="truncate"`? Other options are a dry-run flag that defaults to on that must be explicitly set to false.

Separately, I prefer not to use env variables for configuration of a specific piece of functionality because they're both less visible to developers and because they increase the surface area that can change the behavior of code -- it's not quite, but similar to the principle of locality of behavior: https://htmx.org/essays/locality-of-behaviour

In other words, if we use env variables for configuration of tasks like this, then if I tell you "I ran `make restore-db` locally" you don't know what actually happened; you have to follow up and ask me to tell you what I have TRUNCATE_TABLES set to in my environment. And worse, I might not be able to answer that question -- maybe my env has changed since I ran the command. Whereas if we make configuration like this explicit, I would say "I ran make `restore-db args="truncate"`" and you know exactly what the behavior of that command is -- and we avoid the second issue of needing to know what env vars were set for me at the time I ran the command entirely.

app/src/db/pg_util.py
```diff
@@ -0,0 +1,202 @@
...
+    command = [
+        "psql",
+        "-U",
+        config_dict["user"],
+        "-h",
+        config_dict["host"],
+        "-d",
+        config_dict["dbname"],
+        "-c",
+        "TRUNCATE TABLE alembic_version, user_session, chat_message, document CASCADE;",
```
KevinJBoyer, line 172: Is there a way to do this that avoids explicitly naming the tables? Possibly
```
DO $$ 
DECLARE
    r RECORD;
BEGIN
    FOR r IN (SELECT tablename FROM pg_tables WHERE schemaname = 'public') 
    LOOP
        EXECUTE 'TRUNCATE TABLE public.' || quote_ident(r.tablename) || ' CASCADE';
    END LOOP;
END $$;
```

bit messy but worked locally for me (we might need to inject the schema name)

app/src/db/pg_util.py
KevinJBoyer: I left a longer comment about why an explicit argument might be preferable to using an env var here on the TRUNCATE_TABLES code below!

app/src/db/pg_util.py
```diff
@@ -0,0 +1,202 @@
+import logging
```
KevinJBoyer, line 1: nit: might be better to call this `pg_dump_util.py`

app/tests/src/db/test_pg_util.py
```diff
@@ -0,0 +1,134 @@
...
+
+from src.db import pg_util
+from src.db.models.conversation import ChatMessage, UserSession
+from src.db.models.document import Document
+from tests.src.db.models.factories import ChatMessageFactory, ChunkFactory, UserSessionFactory
+
+
+def _prep_test(monkeypatch, run_command_return_value=True, db_session=None):
+    def _mock_run_command(_command, _stdout_file=None) -> bool:
+        # Don't run any commands b/c they affect the real DB
```
KevinJBoyer, line 17: I think it should be possible to modify `pg_dump` to accept a schema name, and identify the schema name used by the db_client test fixture [here](https://github.com/navapbc/labs-decision-support-tool/blob/main/app/tests/conftest.py#L76), so that these tests actually verify the functionality. LMK if you want to pair on it!

KevinJBoyer (commented):
Nice work on the utility, the code looks great -- I left some nits on configuration and an opportunity to simplify the code by using smart_open. The big change is getting the test coverage set up for this -- I think it should be possible because the test fixtures set up an isolated schema. If you run into roadblocks let me know and we can tackle it together!

yoomlam:
@KevinJBoyer Great points! I'll starting making changes based on the suggestions.
Almost all of the implementation decisions were made based on the speed of me completing the ticket, and the high probability that we may not need to leverage or maintain this feature.

app/src/db/pg_util.py
yoomlam: > I think the intent here is to prevent someone from accidentally truncating with restore-db

Truncation is needed before restoring to avoid duplicate chunks and conflicting `alembic_version` records. I want to give the user a chance to Ctrl-C before clearing the DB.

app/src/db/pg_util.py
yoomlam: At [commit 244337b](https://github.com/navapbc/labs-decision-support-tool/pull/243/commits/244337b4f220838c34258413b45627973f9340ae), I got this error when it's using S3 via `docker compose run -e ENVIRONMENT=dev --rm main-app poetry run pg-dump backup --dumpfile db2.dump`:
```
  File "/app/src/db/pg_dump_util.py", line 141, in _run_command
    result = subprocess.run(command, check=True, stdout=stdout_file)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/subprocess.py", line 992, in __init__
    errread, errwrite) = self._get_handles(stdin, stdout, stderr)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/subprocess.py", line 1725, in _get_handles
    c2pwrite = stdout.fileno()
               ^^^^^^^^^^^^^^^
io.UnsupportedOperation: fileno
```

KevinJBoyer: Boo... I can't think of a way around that, so your original code that writes a temp file first is the way to go. Thanks for trying, sorry for leading you down a rabbit hole!

yoomlam: Got it working using `subprocess.Popen()`!

KevinJBoyer (approved):
LGTM, thank you! I thought about suggesting:

```
def get_schema() -> str
    schema = os.environ.get("DB_SCHEMA", "public")
    if not re.match(r'^[a-zA-Z0-9_]+$', schema):
        logger.fatal("Invalid schema name")
        sys.exit(2)
    return schema
```

but not sure if that's overly paranoid, since you need an AWS IAM role to execute this with sufficient privileges that you could execute arbitrary SQL or commands anyways

app/tests/src/db/test_pg_dump_util.py
```diff
@@ -0,0 +1,120 @@
+import logging
```
KevinJBoyer, line 1: Now that these tests are running the utility directly it'd be easy to add some assertions that validate the behavior vs. logging by executing SELECT queries that validated the counts directly!

app/src/db/pg_dump_util.py
```diff
@@ -0,0 +1,241 @@
...
+            "-h",
+            config_dict["host"],
+            "--schema",
+            schema,
+            config_dict["dbname"],
+        ]
+        return _run_command(command, dumpfile)
+
+
+def _truncate_db_tables(config_dict: dict[str, str], delay_seconds: int) -> bool:
```
KevinJBoyer, line 175: Gah, sorry, I'm too paranoid. Can we bring in a full `db_session` into this function so that we can use a prepared statement instead of running the SQL through the `psql` command, and therefore rule out SQL injection entirely here -- I've got some time this morning so I'll work on a commit to this branch that does this

yoomlam: Feel free to commit and address all your suggestions above.

PR #244, opened 2025-03-04
fix: Allow CORS from prod environment

KevinJBoyer:
## Ticket

n/a

## Changes
 - Update CORS regex for dev instances as well as the prod URL

## Context for reviewers
n/a

## Testing
From an Imagine LA dev environment, confirming the new regex still allows requests:
Screenshot 2025-03-04 at 1 47 48 PM

From an Imagine LA prod environment BEFORE this change:
Screenshot 2025-03-04 at 1 47 29 PM

From an Imagine LA prod environment AFTER the change:
Screenshot 2025-03-04 at 1 47 33 PM

### DIFF
--- a/app/src/app.py
+++ b/app/src/app.py
@@ -11,7 +11,7 @@ app.add_middleware(
     CORSMiddleware,
     # Imagine LA uses port 5173 for development
     allow_origins=["http://localhost:5173"],
-    allow_origin_regex=r"https://dev-social-benefits-navigator-.*\.web\.app",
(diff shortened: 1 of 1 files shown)


fg-nava (approved):

fg-nava:
> for dev instances as well

To clarify, were we still seeing CORS blocking in dev?

KevinJBoyer:
Nope I just wanted to double check that it still worked after I changed it!

PR #246, opened 2025-03-06
refactor: Fix evaluation test files with better test data using existing factory models

fg-nava:
## Changes

* Refactored evaluation metrics tests to use test data that is closer to what we actually can expect 
* Replaced mock objects with actual test data where possible
* using ChunkFactory, DocumentFactory, ChunkWithScore keeping in line with our existing test suite (e.g. factories.py and document.py)

## Context for reviewers

Imrpoves on quality and maintainability of our evaluation metrics and CLI tests

## Testing

* Modified tests passing locally `make test`, `make test-coverage`

### DIFF
--- a/app/tests/src/evaluation/cli/test_cli_evaluate.py
+++ b/app/tests/src/evaluation/cli/test_cli_evaluate.py
@@ -1,28 +1,104 @@
 """Tests for the evaluate CLI module."""
 
 import argparse
+import csv
+import json
+import tempfile
 from pathlib import Path
 from unittest import mock
 
 import pytest
 
+from src.db.models.document import ChunkWithScore
 from src.evaluation.cli import evaluate
+from tests.src.db.models.factories import ChunkFactory, DocumentFactory
 
 
 @pytest.fixture
-def mock_run_evaluation():
-    """Mock the run_evaluation function."""
-    with mock.patch("src.evaluation.cli.evaluate.run_evaluation") as mock_run:
-        yield mock_run
+def mock_git_commit():
+    """Mock git commit hash."""
+    with mock.patch("src.evaluation.metrics.batch.get_git_commit", return_value="test123"):
+        yield
 
 
 @pytest.fixture
-def mock_create_retrieval_function():
+def mock_retrieval_func(test_document):
     """Mock the create_retrieval_function."""
-    with mock.patch("src.evaluation.cli.evaluate.create_retrieval_function") as mock_func:
-        mock_retrieval = mock.MagicMock()
-        mock_func.return_value = mock_retrieval
-        yield mock_func, mock_retrieval
+    retrieval_func = create_mock_retrieval_func(test_document)
+    with mock.patch(
+        "src.evaluation.cli.evaluate.create_retrieval_function", return_value=retrieval_func
+    ):
+        yield retrieval_func
+
+
+@pytest.fixture
+def test_document():
+    """Create a test document with chunks."""
+    document = DocumentFactory.build(
+        name="test_doc",
+        content="Test document content",
+        source="test_dataset",
+        dataset="Imagine LA",  # Match the dataset mapping
+    )
+    chunk = ChunkFactory.build(
+        document=document,
+        content="test chunk content",
+    )
+    document.chunks = [chunk]
+    return document
+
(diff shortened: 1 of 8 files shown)


app/tests/src/evaluation/cli/test_cli_evaluate.py
```diff
@@ -1,28 +1,104 @@
...
 from src.evaluation.cli import evaluate
+from tests.src.db.models.factories import ChunkFactory, DocumentFactory
 
 
 @pytest.fixture
-def mock_run_evaluation():
-    """Mock the run_evaluation function."""
-    with mock.patch("src.evaluation.cli.evaluate.run_evaluation") as mock_run:
-        yield mock_run
+def mock_git_commit():
```
fg-nava, line 18: Currently mocked as we don't currently include `git` as a pypi package in our container so I'm getting the commit hash by doing [`--commit=$(shell git rev-parse HEAD)`](https://github.com/navapbc/labs-decision-support-tool/blob/dbb27a216d3627534ae553135585f88fae11a309/app/Makefile#L301) in Makefile.

Another option is to remove the use of commit tracking throughout this evaluation module as it has limited usefulness.

KevinJBoyer: I had inferred the intent was to let us run the evaluation using the most up-to-date evaluation code but against prior iterations of the retrieval logic -- I agree with your approach here, since we've got the functionality in let's keep it in but keep the testing simple on ourselves!

app/tests/src/evaluation/cli/test_cli_evaluate.py
```diff
@@ -1,28 +1,104 @@
...
+    """Mock git commit hash."""
+    with mock.patch("src.evaluation.metrics.batch.get_git_commit", return_value="test123"):
+        yield
 
 
 @pytest.fixture
-def mock_create_retrieval_function():
+def mock_retrieval_func(test_document):
```
KevinJBoyer, line 25: I think the purpose here is to guarantee (for testing) what chunks get retrieved -- [there's a mock sentence transformer you can use to force similarity](https://github.com/navapbc/labs-decision-support-tool/blob/main/app/tests/mock/mock_sentence_transformer.py#L21) by providing text with different word lengths ("Loquacious vocabulary" will have a higher similarity to "verbose language" than "tiny words", etc.) in the future, which can save mocking an internal function like `create_retrieval_function` in the future

app/tests/src/evaluation/metrics/test_runner.py
```diff
@@ -1,215 +1,295 @@
...
 
 @pytest.fixture
-def mock_questions():
-    """Create mock questions data."""
-    return [
+def mock_datetime():
+    """Mock datetime to return a fixed date."""
+    with patch("src.evaluation.metrics.logging.datetime") as mock_dt:
+        mock_dt.now.return_value = datetime(2024, 1, 1)
+        yield mock_dt
```
KevinJBoyer, line 29: Nice -- if we need this kind of functionality a lot in the future we could consider bringing in https://pypi.org/project/pytest-freezegun/ which I've had good experiences with, but I was surprised to see this was the first time we've needed this functionality so avoiding another dependency for this use case makes sense to me!

KevinJBoyer (approved):
LGTM, great work here! Test coverage looks much improved👨‍🍳

app/tests/src/evaluation/cli/test_cli_evaluate.py
KevinJBoyer: (However FWIW the mock transformer approach felt really clever to me when I first wrote it, but as time has passed I've been less confident that it's the right approach -- if it strikes you that way too I'd love to revisit this and come up with something simpler)

fg-nava: Ah good to know, I wasn't sure initially if this was the right use for the mock sentence transformer. Would like to merge this as is if okay, but I will revisit using MockSentenceTransformer.encode here.

KevinJBoyer: Totally, I don't think you need to update the code here at all.

PR #252, opened 2025-03-11
fix: Use 'app' schema for local DB

yoomlam:
## Ticket

Fix for #243 and https://navalabs.atlassian.net/browse/DST-819

## Changes

[Slack thread](https://nava.slack.com/archives/C06DP498D1D/p1741639200673709): issue: deployments use the schema `app` while local DB uses the schema `public`. Since there is no `app` schema locally, nothing was restored.

Solution approach: make the local DB setup as close to possible as the deployed DB setup (which is generally a good policy) without incurring engineering cost.

## Testing

Download a dump file from S3 `pg_dumps/` folder (prod or dev)
```
make build
make db-recreate
make pg-dump args="restore --dumpfile prod_db-2025-03-10-22_07_35.dump"
```

Make sure the application still works.


app/local.env
```diff
@@ -39,7 +39,7 @@ LOG_ENABLE_AUDIT=FALSE
 DB_HOST=main-db
 DB_NAME=app
 DB_USER=app
-DB_SCHEMA=public
+DB_SCHEMA=app
```
yoomlam, line 42: Since we're no longer using the default `public` schema, `make init-db` needs to set up the schema via (`make init-schema`).

app/pyproject.toml
```diff
@@ -59,6 +59,7 @@ setuptools = "^68.2.2"
 debugpy = "^1.8.1"
 ollama = "^0.2.1"
 jupyter = "^1.1.1"
+pg8000 = "^1.31.2"
```
yoomlam, line 62: Used by `manage.py`

app/Makefile
```diff
@@ -138,6 +138,9 @@ pg-dump:
 alembic_config := ./src/db/migrations/alembic.ini
 alembic_cmd := $(PY_RUN_CMD) alembic --config $(alembic_config)
 
+init-schema:
+	$(PY_RUN_CMD) python src/db/manage.py
```
KevinJBoyer, line 142: nit: could we add this as a script in pyproject.toml for consistency with the other commands?

KevinJBoyer (commented):
Code LGTM, testing it locally now

KevinJBoyer (approved):
Confirmed it works locally!

PR #253, opened 2025-03-11
fix: Update imagine_la scraping; improve refresh-ingestion.sh

yoomlam:
## Ticket

Part of https://navalabs.atlassian.net/browse/DST-849

## Changes

* Fix imagine_la scraping due to [hidden `Learn more` buttons](https://nava.slack.com/archives/C06DP498D1D/p1741722535807229?thread_ts=1741718570.437589&cid=C06DP498D1D)
* Fix and update `app/refresh-ingestion.sh`

## Testing
```sh
./refresh-ingestion.sh imagine_la
```


app/src/ingestion/imagine_la/scrape/scrape_content_hub.py
```diff
@@ -62,10 +62,9 @@
 
     # Expand all the accordions on the page
     # so that we can click into the buttons beneath them
-    accordions = page.locator(".chakra-accordion__button")
-    for accordion_index in range(accordions.count()):
-        accordions.nth(accordion_index).click()
-    page.wait_for_load_state("networkidle")
+    accordions = page.locator(".chakra-accordion__button[aria-expanded='false']")
+    while accordions.count() > 0:
+        accordions.first.click()
```
yoomlam, line 67: Crux of PR. Don't know why this works better but it does.

app/refresh-ingestion.sh
```diff
@@ -161,9 +159,18 @@ EOF
...
 echo_stats(){
     local DATASET_ID="$1"
+    echo "------ $DATASET_ID ----------------"
+    if [ "$DATASET_ID" == "imagine_la" ]; then
+        ls -ld src/ingestion/imagine_la/scrape/pages
+        echo "HTML files scraped: "
+        ls src/ingestion/imagine_la/scrape/pages | wc -l
+    else
+        grep -E 'log_count|item_scraped_count|request_depth|downloader/|httpcache/' "logs/${DATASET_ID}-1scrape.log"
+    fi
```
yoomlam, line 169: A bit of refactoring for `echo_stats` so that it can be called at the end of `all` ingestions to make it easier to copy and paste into slack.

fg-nava (approved):

app/refresh-ingestion.sh
```diff
@@ -16,6 +16,10 @@ ingest_imagine_la() {
     [ "$CONTENTHUB_PASSWORD" ] || { echo "CONTENTHUB_PASSWORD is not set!"; exit 31; }
     # Scrape the website
     make scrape-imagine-la CONTENTHUB_PASSWORD=$CONTENTHUB_PASSWORD 2>&1 | tee logs/${DATASET_ID}-1scrape.log
+    if grep -E 'make:.*Error|log_count/ERROR' "logs/${DATASET_ID}-1scrape.log"; then
+        echo "ERROR: Scraping failed. Check logs/${DATASET_ID}-1scrape.log"
+        exit 33
+    fi
```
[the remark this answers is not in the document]
fg-nava: Addressing error handling in my [comment](https://github.com/navapbc/labs-decision-support-tool/pull/253#issuecomment-2718300281) at the end

yoomlam:
> As feedback, consider refactoring `refresh-ingestion.sh` to python. I see the conveniences of shell but its running our make targets, which are wrapping python anyways. We could directly use these modules instead of shell->make->python indirection.

Good point and good idea! It can definitely be streamlined! Additional considerations:
* Reason for shell script: The shell script originally wrapped running a single dataset and it grew and evolved to running `all` datasets.
* If any part of running `all` datasets fails, we'd want to easily re-run the failing command on the commandline. A shell script makes that easy b/c it's copy and paste from the shell script. If the commands are in Python, there's some impedence to being able to run the failing command (e.g., `zip`, `grep`, `make`) on the commandline.
* This script has a short-term purpose and is (currently) run only by engineers.
* There's other tech debt like making `imagine_la` consistent with the other scraping/ingestion.

PR #254, opened 2025-03-12
fix: LiteralAI exporter: handle threads without citation.source_dataset

yoomlam:
## Ticket

[Slack](https://nava.slack.com/archives/C06DP498D1D/p1741804943062599)

## Changes

For LiteralAI exporter, handle (old) threads without `citation.source_dataset`

## Testing

```
make literalai-exporter args="2024-07-01 2025-03-12"
```

### DIFF
--- a/app/src/evaluation/literalai_exporter.py
+++ b/app/src/evaluation/literalai_exporter.py
@@ -121,7 +121,9 @@ class QARow(NamedTuple):
             program=attribs["benefit_program"] if attribs else None,
             citation_links=("\n".join(c["uri"] for c in citations) if citations else None),
             citation_sources=(
-                "\n".join(f"{c.get('source_dataset', '')}: {c.get('source_name', '')}" for c in citations)
+                "\n".join(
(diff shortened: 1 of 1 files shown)


app/src/evaluation/literalai_exporter.py
```diff
@@ -121,7 +121,9 @@ def from_lai_thread(
             program=attribs["benefit_program"] if attribs else None,
             citation_links=("\n".join(c["uri"] for c in citations) if citations else None),
             citation_sources=(
-                "\n".join(f"{c['source_dataset']}: {c['source_name']}" for c in citations)
+                "\n".join(
+                    f"{c.get('source_dataset', '')}: {c.get('source_name', '')}" for c in citations
+                )
```
fg-nava, line 126: Explaining so I can follow along. These LOC allow for nullable fields in the resulting CSV to support previous export versions that didn't use source_dataset and source_name dating to before last month.

fg-nava (approved):

app/src/evaluation/literalai_exporter.py
yoomlam: Correct

PR #258, opened 2025-03-13
chore: Upgrade to Chainlit 2.4.0

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-858

## Changes

Upgrade to Chainlit 2.4.0

(lint errors are addressed in stacked PR #261 to keep PRs narrowly scoped)

## Testing

```
docker compose up --build main-app
```
Make sure everything still works, i.e., Chainlit and LiteralAI logs.

To enable chat history, authentication must be enabled. Add to `.env`:
```
CHAINLIT_AUTH_SECRET=mysecret
GLOBAL_PASSWORD=mypassword
```
and restart.


app/src/__init__.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+# Affects boto3 and libraries that use it like smart_open
+# Breaking change: https://github.com/boto/boto3/issues/4392
```
yoomlam, line 4: Discovered after a parameterized unit test failed for `s3` but not `local`. Took a while to diagnose a `JSONDecodeError("Extra data", s, end)`.

fg-nava: From [comment](https://github.com/boto/boto3/issues/4435#issuecomment-2648819900) on issue: 

> When uploading a local file to my s3 storage, some kind of checksum seems to be added directory inside the contents of my file: ...

Which file are we uploading to S3 that causes the JSONDecodeError?

app/Makefile
```diff
@@ -308,3 +308,6 @@ literalai-exporter:
 
 literalai-archiver:
 	$(PY_RUN_CMD) literalai-archiver $(args)
+
+literalai-tagger:
+	$(PY_RUN_CMD) literalai-tagger $(args)
```
fg-nava, line 313: Consider moving these commands to dedicated CLI modules (like [generate.py](https://github.com/navapbc/labs-decision-support-tool/blob/main/app/src/evaluation/cli/generate.py) and  [evaluate.py](https://github.com/navapbc/labs-decision-support-tool/blob/main/app/src/evaluation/cli/evaluate.py)) to provide clear argument parsing. This will align with existing CLI pattern in the codebase and improves developer experience.

fg-nava (approved):

PR #265, opened 2025-03-14
feat: Log chats to DB and LiteralAI

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-859

## Changes

* create a Chainlit "Poly" Data Layer that wraps both the LiteralAI data layer (that we currently have) and a Postgres data layer.
* chat messages are logged to both LiteralAI and our own Postgres DB at the same time

What's the impact? This means we don't need to archive LiteralAI logs (or be restricted by LiteralAI limitations) but we can still use LiteralAI's web UI if desired. It also offers additional log analysis possibilities since the data is readily available in our DB.

[Slack](https://nava.slack.com/archives/C06DP498D1D/p1741984624185719)

This PR does not affect behavior for requests via the chatbot API, so this shouldn't affect pilot users.

## Testing

```
make db-migrate
make build
make start
```

Interact with Chainlit chatbot.
Check local DB for corresponding Step and Thread records, noting the UUIDs, which should match those in the LiteralAI web UI.

- Service endpoint: https://p-265-app-dev-1005551137.us-east-1.elb.amazonaws.com
- Deployed commit: 1bfbc0af3c9a5293d015562572baa373735db3e2


app/src/chainlit.py
```diff
@@ -22,7 +23,11 @@
 
 logger = logging.getLogger(__name__)
 
-require_login()
```
yoomlam, line 25: When env vars `GLOBAL_PASSWORD` (and `CHAINLIT_AUTH_SECRET`) are set, `test_require_login_no_password()` fails due to this line running when `import chainlit.config` is executed at the top of the test file. Moving it inside `start()` on line 45 seems to still work.

fg-nava (approved):
I was able to do the migrate and restart the app after rebuilding with the 'asyncpg' dep

app/src/chainlit.py
fg-nava: Confused here. Do we need to set GLOBAL_PASSWORD and CHAINLIT_AUTH_SECRET in .env?

PR #274, opened 2025-03-24
fix: Ensure the `user.id` is consistent across Chainlit data layers

yoomlam:
## Ticket

Part 2 for https://navalabs.atlassian.net/browse/DST-867

## Changes

Ensure the `user.id` is consistent across data layers.
As stated in the code comment:
```
        Unlike other persisted objects (like Thread and Step), the User argument has no id that can be set.
        The ChainlitDataLayer and LiteralDataLayer implementations return a generated UUID user.id.
        Unfortunately, since the UUIDs are different, referencing the user from the Thread consistently
        across data layers is not possible.
        (These Chainlit data layers were implemented to use a single data layer at a time.)
        While the LiteralDataLayer does not allow setting the user.id, the ChainlitDataLayer fortunately does.
        So call LiteralDataLayer.create_user() first and use the generated UUID in the call to
        ChainlitDataLayer.create_user() via User.metadata["uuid"]. In this way, the user.id is the same
        across all data layers.
```

* In `ChainlitPolyDataLayer`, call `LiteralDataLayer.create_user()` first and use the generated UUID in the call to `ChainlitDataLayer.create_user()` so that both data layers have the same `user.id`
* Add class `PostgresDataLayer` to override `ChainlitDataLayer.create_user()` to allow setting `user.id`
* As part of the `archive_threads()`, also download the users so we have all the user data.

## Testing

Added unit test

- Service endpoint: https://p-274-app-dev-891126847.us-east-1.elb.amazonaws.com
- Deployed commit: 9591cb7d0893e074f9f000d8a3926742cf53d607


app/src/chainlit_data.py
```diff
@@ -168,3 +201,47 @@ async def build_debug_url(self) -> str:  # pragma: no cover
...
+    def _get_uuid_metadata(self, user: User) -> str | None:
+        if "uuid" in user.metadata:
+            return user.metadata["uuid"]
+        return None
+
+    async def create_user(self, user: User) -> Optional[PersistedUser]:
+        "Adapted from ChainlitDataLayer.create_user() to use uuid metadata as the id"
+        query = """
+        INSERT INTO "User" (id, identifier, metadata, "createdAt", "updatedAt")
+        VALUES ($1, $2, $3, $4, $5)
```
KevinJBoyer, line 226: My initial thought was to build this using SQLAlchemy so we can take advantage of the typing, but it looks like SQLAlchemy doesn't support the $1 binding syntax :/

app/src/chainlit_data.py
```diff
@@ -168,3 +201,47 @@ async def build_debug_url(self) -> str:  # pragma: no cover
...
+        SET metadata = $3
+        RETURNING *
+        """
+        now = await self.get_current_timestamp()
+        params = {
+            "id": self._get_uuid_metadata(user) or str(uuid.uuid4()),
+            "identifier": user.identifier,
+            "metadata": json.dumps(user.metadata),
+            "created_at": now,
+            "updated_at": now,
```
KevinJBoyer, line 237: I think you could leave these out since they have defaults of `now()`

KevinJBoyer (approved):
LGTM, thank you!

app/src/chainlit_data.py
yoomlam: This function was copied from `ChainlitDataLayer` -- only line 233 was changed. I'd like to keep the function the same to easily update it if needed.

app/src/chainlit_data.py
yoomlam: I'll change this function to call `super(). create_user()`

KevinJBoyer: 👍 OK, no worries, your explanation of using the same code from the CL implementation makes sense to me!

PR #283, opened 2025-04-01
fix: Handle empty NavigableString in beautifulsoup 4.13

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-877

## Changes
 - Explicitly bring in beautifulsoup4 as a dependency
 - Adjust typing to match

## Context for reviewers
 - Previously we were inadvertently bringing in beautifulsoup4 via another library's dependency on it
 - This was upgraded to 4.13, which caused some typing to change and a minor behavior change (bringing in empty NavigableStrings where previously they were omitted)

## Testing
Run `./refresh-ingestion.sh la_policy` and confirm no exceptions (we only strip NavigableString if it is whitespace)

- Service endpoint: https://p-283-app-dev-268913096.us-east-1.elb.amazonaws.com
- Deployed commit: aa077d9a105ec2ecab860d2660aff7da1143630f


app/src/ingestion/scrapy_dst/spiders/la_policy_spider.py
```diff
@@ -983,6 +992,16 @@ def _remove_empty_elements(contents: list[PageElement]) -> None:
...
 
 
+def _strip_empty_strings(elements: list[PageElement]) -> list[PageElement]:
+    # Contents should either be empty string or Tag
+    for child in elements:
+        if isinstance(child, NavigableString) and child.strip() == "":
+            continue
+        assert isinstance(child, Tag)
+
+    return [child for child in elements if isinstance(child, Tag)]
```
KevinJBoyer, line 1002: The typing looks a bit weird here (we say we're returning list[PageElement] but we know that we're returning only list[Tag]) but we want to be able to assign this back to `.contents` which is of type `list[PageElement]`.

yoomlam (approved):
Code changes look good

PR #282, opened 2025-03-31
feat: Merge contiguous cited subsections

yoomlam:
## Ticket

Part of https://navalabs.atlassian.net/browse/DST-646

## Changes

Sequential contiguous subsections that are cited together are merged into a single citation.

Also:
- fix bullet rendering by ensuring a new line occurs before any bullet character.
- fix extra spaces after citation heading

## Testing

image
Citation #3 merges 3 citations/paragraphs:
image

## Preview environment for app
- Service endpoint: https://p-282-app-dev-1134576066.us-east-1.elb.amazonaws.com
- Deployed commit: 8dda71b2fade3f0a5f220154301b22586c976d75

PR #285, opened 2025-04-02
refactor: Use simplify_citation_numbers() consistently

yoomlam:
## Ticket

Address tech debt while working around the same code as #282 

## Changes

Use `simplify_citation_numbers()` in `format.py` for consistency. 

## Testing

No change in behavior

- Service endpoint: https://p-285-app-dev-1276489056.us-east-1.elb.amazonaws.com
- Deployed commit: e8fa09c0df2ed9b1873248fe4e575ee1b37fecad


app/src/format.py
```diff
@@ -7,12 +7,7 @@
 
 import markdown
 
-from src.citations import (
-    CITATION_PATTERN,
-    merge_contiguous_cited_subsections,
-    move_citations_after_punctuation,
-    remap_citation_ids,
```
yoomlam, line 14: Instead of calling this three functions, call `simplify_citation_numbers()` instead.

app/src/format.py
```diff
@@ -194,38 +179,30 @@ def _get_breadcrumb_html(headings: Sequence[str] | None, document_name: str) ->
 _footnote_index = 0
 
 
-# FIXME: Refactor to reduce code replication with replace_citation_ids();
-#        perhaps use simplify_citation_numbers() like chat_api and batch_process
```
yoomlam, line 198: This PR addresses this `FIXME` tech debt.

app/src/format.py
```diff
@@ -194,38 +179,30 @@ def _get_breadcrumb_html(headings: Sequence[str] | None, document_name: str) ->
...
+
+    # Replace (citation-<index>) with a superscript citation link to the respective accordion item
     def replace_citation(match: Match) -> str:
-        citation_id = match.group(1)
-        # Remove citation for chunks that don't exist alone
-        if citation_id not in remapped_citations:
-            logger.error(
-                "LLM generated a citation for a reference (%s) that doesn't exist.", citation_id
-            )
-            return ""
```
yoomlam, line 217: This is already done in `replace_citation_ids()`

app/src/format.py
```diff
@@ -194,38 +179,30 @@ def _get_breadcrumb_html(headings: Sequence[str] | None, document_name: str) ->
...
-
-        chunk = remapped_citations[citation_id].chunk
+        citation_id = match.group(1).removeprefix("citation-")
+
```
fg-nava, line 197: Where do you handle the case when citation_id not in citation_dict?

yoomlam: Before this `_add_citation_links()` function is called, `simplify_citation_numbers()` is called. `simplify_citation_numbers()` calls `replace_citation_ids()`, which removes unknown citation ids, so `citation_id` is always in `citation_dict` (unless there's a bug).

fg-nava (approved):

PR #289, opened 2025-04-07
feat: Add AWS Bedrock's Anthropic Claude 3.7 LLM

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-301

## Changes

For Chainlit only: Enable using LLMs from AWS Bedrock, i.e. Anthropic Claude 3.7 LLM

## Testing
* For local testing, ensure `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env vars are set for the container running Chainlit.
* Or just go to https://p-289-app-dev-1929265981.us-east-1.elb.amazonaws.com/chat/

Then, in Chainlit settings UI, pick `Bedrock Claude 3.7 Sonnet` and enable `Show message-assessment attributes`.
- Ask a question like `what can you tell me?` to demonstrate the `canned_response` is set.
- Ask a question that doesn't set the `canned_response` like `Tell me about In-Home Supportive Services`
- Ask a follow-up question

♻️ Environment destroyed ♻️

### DIFF
--- a/infra/modules/service/bedrock.tf
+++ b/infra/modules/service/bedrock.tf
@@ -21,7 +21,6 @@ data "aws_iam_policy_document" "bedrock" {
 }
 
 resource "aws_iam_role_policy_attachment" "bedrock" {
-  count      = 1
   role       = aws_iam_role.app_service.name
   policy_arn = aws_iam_policy.bedrock.arn
 }


app/src/generate.py
```diff
@@ -34,6 +45,19 @@ def get_models() -> dict[str, str]:
     return models
 
 
+def _has_aws_access() -> bool:
+    # LiteLLM requires these env variables to access Bedrock models - https://docs.litellm.ai/docs/providers/bedrock
+    if "AWS_ACCESS_KEY_ID" not in os.environ or "AWS_SECRET_ACCESS_KEY" not in os.environ:
+        return False
+    if "AWS_REGION" not in os.environ and "AWS_DEFAULT_REGION" in os.environ:
+        os.environ["AWS_REGION"] = os.environ["AWS_DEFAULT_REGION"]
```
yoomlam, line 53: @KevinJBoyer Deployed app at https://p-289-app-dev-1929265981.us-east-1.elb.amazonaws.com/chat doesn't show the Bedrock LLM. Which [LiteLLM Bedrock authentication](https://docs.litellm.ai/docs/providers/bedrock#boto3---authentication) method should we be using?

yoomlam: Should we do [this](https://github.com/BerriAI/litellm/blob/55a20c7cce99a93d36a82bf3ae90ba3baf9a7f89/litellm/llms/bedrock_httpx.py#L345) to get the `aws_access_key_id` and `aws_secret_access_key`?

KevinJBoyer: Sorry, only saw your message ("Should we do..") after sending that ^ comment. Looking now

app/src/generate.py
```diff
@@ -23,6 +25,15 @@ def get_models() -> dict[str, str]:
         models |= {"OpenAI GPT-4o": "gpt-4o"}
     if "ANTHROPIC_API_KEY" in os.environ:
         models |= {"Anthropic Claude 3.5 Sonnet": "claude-3-5-sonnet-20240620"}
+    if _has_aws_access():
+        # If you get "You don't have access to the model with the specified model ID." error,
+        # remember to request access to Bedrock models ...aws.amazon.com/bedrock/home?region=us-east-1#/modelaccess
```
yoomlam, line 30: @KevinJBoyer For local development, I requested access to the Bedrock model via AWS Console's web UI. How should this be done for the deployed app?

app/src/generate.py
yoomlam: How does `smart_open()` get access to S3 without us having to set up an IAM role?

KevinJBoyer: Oh right, we're already using AWS stuff... let me check the container config in the console.

yoomlam: https://docs.litellm.ai/docs/providers/bedrock#sts-role-based-auth
Where should I get aws_role_name and aws_session_name?

KevinJBoyer: This is the actual auth flow we would ideally use, which is now marked deprecated, under `Create a client from AWS profile in ~/.aws/config:`:
https://docs.litellm.ai/docs/providers/bedrock#passing-an-external-bedrockruntimeclient-as-a-parameter---completion

I'm not sure what to make of that "experimental" note. My read is that LiteLLM doesn't actually use the boto3 client. I wonder if we could pass aws_role_name and aws_session_name as env vars to the container?

app/src/generate.py
KevinJBoyer: I think your access requested applied to the whole account (I see "Claude 3.7 access granted" when I got to bedrock under my IAM role), so it should work in deployed envs as well. I don't think there's a way around it being a manual part of the set up process if we deployed the app to a different AWS account though

app/src/generate.py
KevinJBoyer: Following up here, it looks like LiteLLM doesn't need any special configuration to assume the IAM role:
`completion(model="bedrock/us.anthropic.claude-3-7-sonnet-20250219-v1:0", messages=[{"role":"system", "content":"Test message"},{"role":"user", "content":"Test message."}])` yields:

```litellm.exceptions.APIConnectionError: litellm.APIConnectionError: BedrockException - {"Message":"User: arn:aws:sts::6....5:assumed-role/app-dev-app/4d761d....ee494 is not authorized to perform: bedrock:InvokeModel on resource: arn:aws:bedrock:us-east-1:65....45:inference-profile/us.anthropic.claude-3-7-sonnet-20250219-v1:0 because no identity-based policy allows the bedrock:InvokeModel action"}```

So I think the next step is to provide it the correct BedRock permissions. I think we would add another file, like bedrock.tf, to services, to define these permissions. Here's an example file: https://github.com/navapbc/labs-decision-support-tool/blob/kb/skip-empty-json/infra/modules/service/command_execution.tf#L1

And this link has the correct IAM permissions to specify (I think) -- https://github.com/BerriAI/litellm/discussions/8949

Comment (KevinJBoyer) [reply app/src/generate.py:53]: I picked `command_execution.tf` because it looked like the simplest one, though note that that particular file has a quirk in that it dynamically attaches the policy or not (`count      = var.enable_command_execution ? 1 : 0`) which we wouldn't need to replicate here

File: app/src/chainlit.py [hunk 3]
```diff
@@ -24,7 +24,6 @@
 logger = logging.getLogger(__name__)
 
 
-# TODO: Update chat_api.py to use ChainlitPolyDataLayer
```
yoomlam, line 27: Unrelated to this PR: remove completed TODO

infra/modules/service/bedrock.tf
```diff
@@ -0,0 +1,27 @@
...
+    actions = [
+      "bedrock:InvokeModel",
+      "bedrock:InvokeModelWithResponseStream"
+    ]
+    resources = ["*"]
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "bedrock" {
+  count      = 1
```
KevinJBoyer, line 24:
```suggestion
```

KevinJBoyer (approved):
LGTM!

PR #292, opened 2025-04-14
feat: Import Benefit Hub content from Contentful API

KevinJBoyer:
## Ticket

https://navalabs.atlassian.net/browse/DST-917

## Changes
 - Update scrape_content_hub.py -> scrape.py to use the Contentful Python SDK instead of scraping HTML pages
 - Get rid of the imagine_la/scrape folder since we no longer need a contained script with playwright, etc.
 - Update ingest.py to use the generated .html from scrape_content_hub.py (instead of the content hub webpages directly)

## Context for reviewers
 - Had to add a heuristic (checking for "test" at the end of the URL) for unpublished test content because this isn't reported in the Contentful API

## Testing
Set CONTENT_HUB_ACCESS_TOKEN and CONTENT_HUB_SPACE_ID in your env (see 1Password for these values), then:
`./refresh-ingestion.sh imagine_la` and inspect DB to see the ingested chunks:
image
And check locally:
image

- Service endpoint: https://p-292-app-dev-260269913.us-east-1.elb.amazonaws.com
- Deployed commit: 4e3f4110990118f394d8e75b8e277f1ffd6dcee5


docs/data-management.md
fg-nava, line 1: We should update this doc with instructions to set the new Contentful env vars for Imagine LA. As well as update the current CLI command `make scrape-imagine-la CONTENTHUB_PASSWORD=your_password_here`

fg-nava (approved):
Nice to be able to simplify Imagine LA extraction to use the Contentful API over Playwright. Just left a comment for an early check on the environment variable in the refresh shell script and suggestion to update the docs with instructions on how to access and set the Contenful token and space ID. 

I was able to run the new build with no issues.

PR #293, opened 2025-04-16
feat: Add streaming endpoint with citations and traces

fg-nava:
## Ticket  
https://navalabs.atlassian.net/browse/DST-846  

## Changes  
- Added, SSE streaming endpoints: `/api/query_init` and `/api/query_stream`  
- `QueryInitRequest` / `QueryInitResponse` models and `QueryInit` route 
- Factored out `_prepare_messages()` in `generate.py` and implemented `generate_streaming_async`  
- Modified `run_query()` to accept a `streaming` flag and separate streaming vs. non‑streaming logic  
- Extended `ChatEngineInterface` and `BaseEngine` with `on_message_streaming` methods  
- Updated `ImagineLaEngine` to handle streaming, canned responses, and alert formatting  
- In `query_stream`, implemented async SSE generator emitting `alert`, `chunk`, and `done` events  

## Context for reviewers  
This PR sets up streamed chat handling via Server-Sent Events (SSE) as outlined in our [ADR](https://github.com/navapbc/labs-decision-support-tool/blob/c0cc3b5222d121b551aea39d39a816b9c7afefa3/docs/decisions/2025-03-31-streaming-responses-for-chatbot.md):
1. Initialization (`/query_init`): creates a message ID and persists the question.  
2. Streaming (`/query_stream`): uses SSE to emit:  
   - `alert` (policy notices)  
   - `chunk` (partial LLM responses)  
   - `done` (final response with citations and metadata)  

Streaming reuses existing engine logic via `generate_streaming_async()` and the updated `run_query(streaming=True)` path. Engines return an async generator plus message attributes and subsections. Alert messages are prefixed exactly once, then chunks follow.  

## Testing  
- SSE endpoints: `test_query_stream_basic`, `test_query_stream_with_alert`  
- Chat engine streaming: streaming canned responses, context‑based streaming (`test_chat_engine.py`)  
- LLM streaming: `test_generate_streaming_async()`
- Refer to this PR for testing a sandbox client UI demonstrating streaming: https://github.com/navapbc/labs-decision-support-tool/pull/296

- Service endpoint: https://p-293-app-dev-1687611467.us-east-1.elb.amazonaws.com
- Deployed commit: 2e88b9f7243c7571e5c7f5edd5f270ebb36ab28a


app/src/chat_api.py
```diff
@@ -480,7 +495,150 @@ async def process_request() -> tuple[QueryResponse, StepDict]:
...
+                    full_response += chunk
+                    yield {"event": "chunk", "data": chunk}
+
+                # Process final response with citations
+                query_response, meta = await run_query(
+                    engine, question_content, chat_history, streaming=True
+                )
+
+                # Send the remapped final response so client can update displayed text
+                yield {"event": "remapped_response", "data": query_response.response_text}
```
fg-nava, line 605: Citation remapping ("remapped_response") happens in event sent before the "done" event

app/src/chat_engine.py
```diff
@@ -451,3 +522,34 @@ def on_message(
             return self._build_response_with_context(question, attributes, chat_history)
 
         return self._build_response(question, attributes, chat_history)
+
+    async def on_message_streaming(
```
fg-nava, line 526: `on_message_streaming` extended to include ImagineLA specific features like alert message and canned response

app/src/chat_api.py
```diff
@@ -498,27 +656,54 @@ def _validate_chat_history(request: QueryRequest, chat_history: ChatHistory) ->
 
 
 async def run_query(
```
fg-nava, line 658: extended `run_query` to include a `streaming` boolean param

app/src/chat_engine.py
```diff
@@ -64,6 +65,7 @@ def __init__(
 class ChatEngineInterface(ABC):
     engine_id: str
     name: str
+    llm: str = "gpt-4o"  # Default LLM to use
```
yoomlam, line 68: I suppose you can remove `llm` from `BaseEngine` since you added `llm` here.

app/src/chat_engine.py
```diff
@@ -150,6 +158,20 @@ def on_message(
...
+    async def on_message_streaming(
+        self, question: str, chat_history: Optional[ChatHistory] = None
+    ) -> tuple[AsyncGenerator[str, None], MessageAttributes, Sequence[Subsection]]:
+        # Start timing system_prompt_1
+        start_time = time.perf_counter()
+        attributes = analyze_message(self.llm, self.system_prompt_1, question, MessageAttributes)
+        system_prompt_1_duration = time.perf_counter() - start_time
+        logger.info(
+            f"System Prompt 1 (analyze_message) took {system_prompt_1_duration:.2f} seconds"
+        )
```
yoomlam, line 170: This block duplicates code from `on_message()`. Extract into a reusable function so that any updates only need to be made in one location.

app/src/chat_engine.py
```diff
@@ -218,6 +240,55 @@ def _build_response_with_context(
...
+    async def _build_streaming_response(
+        self,
+        question: str,
+        attributes: MessageAttributes,
+        chat_history: Optional[ChatHistory] = None,
+    ) -> tuple[AsyncGenerator[str, None], MessageAttributes, Sequence[Subsection]]:
+        """Helper method to build a streaming response with or without context"""
+        subsections: Sequence[Subsection] = []
+
+        if attributes.needs_context:
```
yoomlam, line 252: Refactor this big function into something like https://github.com/navapbc/labs-decision-support-tool/blob/17df6e263537b867ed1dec280e6117ca40e97aa9/app/src/chat_engine.py/#L148-L151

app/tests/mock/mock_completion.py
```diff
@@ -11,4 +12,17 @@ def mock_completion(model, messages, **kwargs):
...
+
+    # If streaming is requested, return chunks instead
+    if kwargs.get("stream", False):
+        # Split into small chunks while preserving spaces
+        chunk_size = 10
+        chunks = []
+        for i in range(0, len(mock_response), chunk_size):
+            chunk = mock_response[i : i + chunk_size]
+            chunks.append(
+                SimpleNamespace(choices=[SimpleNamespace(delta=SimpleNamespace(content=chunk))])
```
yoomlam, line 24: TIL: `SimpleNamespace`

yoomlam (commented):
If we keep both streaming and non-streaming functionality, we should refactor (in another PR) so there is less code duplication, perhaps by putting most of the code in streaming functions, then wrapping streaming functions with a non-streaming function counterpart.
Alternatively, if we pick one, we should pick streaming since it's easier to adapt code to be non-streaming than going the other way. Lesson learned: default to streaming capability.

fg-nava:
> If we keep both streaming and non-streaming functionality, we should refactor (in another PR) so there is less code duplication, perhaps by putting most of the code in streaming functions, then wrapping streaming functions with a non-streaming function counterpart.
> 
> Alternatively, if we pick one, we should pick streaming since it's easier to adapt code to be non-streaming than going the other way. Lesson learned: default to streaming capability.

Yes, I refrained from extending existing functions (aside from run_query) in this round. Agree lot of space to deduplicate.

yoomlam (approved):
Tested via http://localhost:8000/api/streaming-test in #296. Nice work!
I look forward to Kevin's feedback on this PR.

app/src/chat_api.py
```diff
@@ -480,7 +495,150 @@ async def process_request() -> tuple[QueryResponse, StepDict]:
...
+    request: Request, id: str, user_id: str, session_id: str
+) -> EventSourceResponse:
+    """
+    Streams the response for a query initiated with query_init.
+    Uses Server-Sent Events (SSE) to stream the response chunks to the client.
+    """
+    with db_session_context_var() as db_session:
+        # Get session information
+        session = await _init_chat_session(user_id, session_id, new_session=False)
+        chat_history = _load_chat_history(session.user_session)
```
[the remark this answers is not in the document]
fg-nava: Repurposed _validate_chat_history to allow for a new_session bool param. This allows us to remove the extra DB request for chat history in query_init

app/src/chat_engine.py
fg-nava: Bookmarking for follow-up PR

app/src/chat_engine.py
fg-nava: Bookmarking for follow-up PR

PR #298, opened 2025-04-25
feat: Add streaming to chainlit chat UI

fg-nava:
## Ticket
https://navalabs.atlassian.net/browse/DST-946

## Changes
- Changed `on_message` in chainlit.py to use our new streaming functionality:
  - `on_message_streaming()` to get async response generator
  - Streams tokens in with `msg.stream_token()`
  - `msg.update()` to get final message with remapped citations and formatting

## Context for reviewers
Adds streaming support to our Chainlit UI interface
Works by:
1. Creating empty message first then streams tokens
2. After streaming is done, updates message with remapped and formatted citations
3. Reuses all existing functionality for the citations remapping, metadata, and tthe attributes

Chainlit docs:
- [Streaming Messages](https://docs.chainlit.io/api-reference/message#stream-a-message)
- [Updating Messages](https://docs.chainlit.io/api-reference/message#update-a-message)

## Testing
Tested functionality through `/chat` UI, verified threads in LiteralAI logs:
- Tested with question: "What benefits are available for low-income families in Los Angeles?"
- Tested with "batch processing"
- Tested canned response with out-of-scope question

Did not change unit tests needed as this modifies only the UI layer using streaming functionality from previous PR.

- Service endpoint: https://p-298-app-dev-1217013430.us-east-1.elb.amazonaws.com
- Deployed commit: 3a26eb60d117109d09f25c2bd83fe077bb9117ea

### DIFF
--- a/app/src/chainlit.py
+++ b/app/src/chainlit.py
@@ -254,21 +254,13 @@ async def on_message(message: cl.Message) -> None:
         # Update the message with formatted content and metadata
         # See: https://docs.chainlit.io/api-reference/message#update-a-message
         msg.content = msg_content
-        msg.metadata = {
-            "system_prompt": engine.system_prompt_2,
-            "subsections": [
-                {
-                    "id": citations.id,
-                    "chunk.id": str(citations.chunk.id),
-                    "document.name": citations.chunk.document.name,
-                    "headings": citations.text_headings,
-                    "text": citations.text,
-                }
-                for citations in final_result.subsections
-            ],
-            "raw_response": final_result.response,
-            "attributes": attributes.model_dump(),
-        }
+        result = OnMessageResult(
+            response=final_result.response,
+            system_prompt=engine.system_prompt_2,
+            attributes=attributes,
+            subsections=final_result.subsections,
+        )
+        msg.metadata = _get_retrieval_metadata(result)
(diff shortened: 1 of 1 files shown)


app/src/chainlit.py
```diff
@@ -227,29 +225,54 @@ async def on_message(message: cl.Message) -> None:
...
+                    "chunk.id": str(citations.chunk.id),
+                    "document.name": citations.chunk.document.name,
+                    "headings": citations.text_headings,
+                    "text": citations.text,
+                }
+                for citations in final_result.subsections
+            ],
+            "raw_response": final_result.response,
+            "attributes": attributes.model_dump(),
+        }
```
yoomlam, line 271: Why not reuse or update `_get_retrieval_metadata()`, which returns an `chunks` dict entry?

fg-nava: Good call, resolved 🕶️

yoomlam (approved):
Thanks! 
Note that if a canned response is triggered, it does not stream the response.

fg-nava:
> Thanks! Note that if a canned response is triggered, it does not stream the response.

Good catch, canned response isn't streamed from `/query_stream` endpoint either. Will fix in next PR.

PR #301, opened 2025-04-30
fix: Improve referral link decision criteria in ImagineLaEngine

fg-nava:
## Ticket

https://navalabs.atlassian.net/browse/DST-951

## Changes

- Updated `system_prompt_1` in ImagineLaEngine to improve referral link decision criteria
- Added specific decision criteria for when to use referral links
- Examples to illustrate correct usage

## Context for reviewers

Message analysis prompt was incorrectly using referral links for questions about foster youth resources, where it should have instead searched the knowledge base. The previous prompt instruction was previously: "If the user's question is about a referral link below...". This appeared to lead to false positives where LLM serves an irrelevant link for foster youth resource questions.

Proposed edits makes criteria explicit and defaults to `need_context=True` when uncertain. [Slack link](https://nava.slack.com/archives/C06DP498D1D/p1746041077878979)

## Testing

Tested with problematic queries that previously triggered incorrect referral links:

Query: "Where can I find foster youth resources?"
- Before: LGBTQ Resources link
- After: Searches and returns foster youth benefits info

- Service endpoint: https://p-301-app-dev-1808349357.us-east-1.elb.amazonaws.com
- Deployed commit: e271d60963b36b530d30a6bb1e594790eb70c7aa

### DIFF
--- a/app/src/chat_engine.py
+++ b/app/src/chat_engine.py
@@ -403,7 +403,9 @@ If the user's question is about these questions related to the benefit navigator
 then set canned_response to: "To get support with that issue, select "Need help? Contact the support team" at the top of this chatbot to open a ticket with the operations team. You can also email us at [socialbenefithelp@imaginela.org](mailto:socialbenefithelp@imaginela.org)"
 
 For referral links below, only set canned_response to a referral link if:
-The user is explicitly asking how to obtain/access/find that specific resource (e.g., "How do I get an ID card?") and the question is narrowly about the process of obtaining that resource and the question does not appear to be about eligibility, benefits programs, or resources for specific populations
+- Sser is explicitly asking how to obtain/access/find that specific resource (e.g., "How do I get an ID card?")
+- Question is narrowly about the process of obtaining that resource
+- Question does not appear to be about eligibility, benefits programs, or resources for specific populations
 
 If these criteria are met, then set canned_response to:
 "Here's a trusted link to learn more: [referral link title](referral link). \
--- a/app/src/chat_engine.py
+++ b/app/src/chat_engine.py
@@ -402,13 +402,8 @@ If the user's question is about these questions related to the benefit navigator
 - Or other kinds of support questions for the Benefit Navigator tool
 then set canned_response to: "To get support with that issue, select "Need help? Contact the support team" at the top of this chatbot to open a ticket with the operations team. You can also email us at [socialbenefithelp@imaginela.org](mailto:socialbenefithelp@imaginela.org)"
 
-If canned_response is not set by the above instructions, then consider referral links. 
-Only use a referral link if the user is explicitly asking how to obtain/access/find the exact resource described by one of the referral links below.
-
 For referral links below, only set canned_response to a referral link if:
-1. The user is explicitly asking how to obtain/access/find that specific resource (e.g., "How do I get an ID card?")
-2. and the question is narrowly about the process of obtaining that resource
-3. and the question does not appear to be about eligibility, benefits programs, or resources for specific populations
+The user is explicitly asking how to obtain/access/find that specific resource (e.g., "How do I get an ID card?") and the question is narrowly about the process of obtaining that resource and the question does not appear to be about eligibility, benefits programs, or resources for specific populations
 
 If these criteria are met, then set canned_response to:
 "Here's a trusted link to learn more: [referral link title](referral link). \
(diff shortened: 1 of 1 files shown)


app/src/chat_engine.py
```diff
@@ -428,6 +445,13 @@ class ImagineLaEngine(BaseEngine):
 - [Find LADWP contact info](https://www.ladwp.com/account/customer-service/customer-service-centers)
 - [Find Legal Aid](https://lafla.org/get-help/)
 - [See Federal Poverty Levels](https://aspe.hhs.gov/topics/poverty-economic-mobility/poverty-guidelines)
+- [Find Diapers](https://www.phfewic.org/en/diaper-resources-in-la-county)
+
+# Examples to illustrate correct referral link decisions:
+# Question: "How do I get an ID card?" → Use referral link for "Get an ID card"
```
yoomlam, line 451: Since I'm unsure if LLMs can interpret the arrow, I'd change it to an `if... then` statement
```suggestion
# If the user asks "How do I get an ID card?", then use referral link for "Get an ID card"
```

app/src/chat_engine.py
```diff
@@ -402,13 +402,30 @@ class ImagineLaEngine(BaseEngine):
...
+# 1. Coverage list of supported benefit programs
+# 2. Password reset instructions for the Benefit Navigator
+# 3. Benefit Navigator support issue instructions
+# Only if none of the above apply should you consider referral links
+
+# Referral links section
+# Important: Be strict about when to use referral links
+# Only use a referral link if the user is explicitly asking how to obtain/access/find the exact resource described by one of the referral links below.
+# By default, prefer setting needs_context=True to search the knowledge base.
+# When uncertain, do not use referral links - set needs_context=True instead.
```
yoomlam, line 415: Is the `# ` prefix needed?

app/src/chat_engine.py
```diff
@@ -402,13 +402,30 @@ class ImagineLaEngine(BaseEngine):
 - Or other kinds of support questions for the Benefit Navigator tool
 then set canned_response to: "To get support with that issue, select "Need help? Contact the support team" at the top of this chatbot to open a ticket with the operations team. You can also email us at [socialbenefithelp@imaginela.org](mailto:socialbenefithelp@imaginela.org)"
 
-If the user's question is about a referral link below, set canned_response to: \
+# Precedence of canned_response:
+# 1. Coverage list of supported benefit programs
+# 2. Password reset instructions for the Benefit Navigator
+# 3. Benefit Navigator support issue instructions
+# Only if none of the above apply should you consider referral links
```
yoomlam, line 409: Can this be simplified to:
```suggestion
If canned_response is not set by the above instructions, then consider referral links
```

app/src/chat_engine.py
```diff
@@ -402,13 +402,30 @@ class ImagineLaEngine(BaseEngine):
...
+
+# Referral links section
+# Important: Be strict about when to use referral links
```
yoomlam, line 412: Why is this needed? I'm not sure how to interpret this instruction myself. The system prompt instructions should already be strict.

app/src/chat_engine.py
```diff
@@ -402,13 +402,30 @@ class ImagineLaEngine(BaseEngine):
...
+# Only use a referral link if the user is explicitly asking how to obtain/access/find the exact resource described by one of the referral links below.
+# By default, prefer setting needs_context=True to search the knowledge base.
```
yoomlam, line 414: The default for `needs_context` is already set to `True` on line 358. Why is it also needed here?

app/src/chat_engine.py
```diff
@@ -402,13 +402,30 @@ class ImagineLaEngine(BaseEngine):
...
+# When uncertain, do not use referral links - set needs_context=True instead.
+
+For referral links below, only set canned_response to a referral link if:
+1. The user is explicitly asking how to obtain/access/find that specific resource (e.g., "How do I get an ID card?")
+2. and the question is narrowly about the process of obtaining that resource
+3. and the question does not appear to be about eligibility, benefits programs, or resources for specific populations
```
yoomlam, line 420: Consolidate all the conditionals into a single block of text, i.e., incorporate `if the user is explicitly asking how to obtain/access/find the exact resource described by one of the referral links below.` (from line 413), then incorporate with my earlier suggestion `If canned_response is not set by the above instructions, then consider referral links`.

yoomlam (commented):
The new logic is hard for me to follow; let's try to make it more straight-forward so that future edits around this aren't too convoluted.

app/src/chat_engine.py
yoomlam: Looks like this is still the same. Is this intentional?

yoomlam (approved):
Great!
The only open issue in my mind is https://github.com/navapbc/labs-decision-support-tool/pull/301#discussion_r2069405389

PR #310, opened 2025-05-14
ci: Extract evaluation ID from share URL

fg-nava:
## Changes
- Fixed evaluation ID extraction in GH Action to capture eval ID from promptfoo output URL

## Context
Previously the workflow was trying to extract the eval ID from the JSON output file, but Promptfoo actually provides it in the CLI output URL. Changd script to capture and parse this URL instead.

## Testing
- Tested by running the PR=true workflow - confirmed results are now being exported to the Google Sheet.
- Need to test it correctly exports results to Google Sheets when running manually via Actions UI (only available in main branch)

Attempted to use `act workflow_dispatch` for local testing of a non-pull request event but still getting errors and haven't had luck debugging

- Service endpoint: https://p-310-app-dev-652842717.us-east-1.elb.amazonaws.com
- Deployed commit: de7e268ca702b4b357d459ee39ea6de40421c63f

### DIFF
--- a/.github/workflows/promptfoo-googlesheet-evaluation.yml
+++ b/.github/workflows/promptfoo-googlesheet-evaluation.yml
@@ -10,6 +10,7 @@ on:
       - 'app/src/generate.py'
       - 'app/promptfoo/promptfooconfig.ci.yaml'
       - 'app/promptfoo/generateUniqueId.js'
+      - '.github/workflows/promptfoo-googlesheet-evaluation.yml'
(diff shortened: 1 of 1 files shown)


KevinJBoyer (approved):
It works! I should have done this on your earlier PR rather than just reading the code -- If you go to the action, there's a drop down for "Use workflow from..." where you can select the PR branch. I ran the workflow in this branch against the test sheet and confirm it ran and wrote to a separate tab!

PR #321, opened 2025-06-17
fix: Use singleton PostgresDBClient (Sqlalchemy engine)

yoomlam:
## Ticket

https://navalabs.atlassian.net/browse/DST-1025
Test changes for https://navalabs.atlassian.net/browse/DST-1042

### DIFF
--- a/.github/workflows/promptfoo-googlesheet-evaluation.yml
+++ b/.github/workflows/promptfoo-googlesheet-evaluation.yml
@@ -118,9 +118,9 @@ jobs:
           EVAL_OUTPUT_FILE="/tmp/promptfoo-output.txt"
 
           if [ -n "$PROMPTFOO_API_KEY" ]; then
-            promptfoo eval --max-concurrency 1 --config "/tmp/promptfooconfig.processed.yaml" --share --output "${OUTPUT_JSON_FILE}" --no-cache | tee "${EVAL_OUTPUT_FILE}"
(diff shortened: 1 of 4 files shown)


yoomlam:
Promptfoo run with Gemini 2.5 pro: https://github.com/navapbc/labs-decision-support-tool/actions/runs/15737577700

yoomlam:
DB connections spiked but went down: 
image

app/tests/src/test_retrieve.py
```diff
@@ -86,7 +86,7 @@ def test_retrieve_with_scores(app_config, db_session, enable_factory_create):
     results = retrieve_with_scores("Very tiny words.", retrieval_k=2, retrieval_k_min_score=0.0)
 
     assert len(results) == 2
-    assert results[0].chunk == short_chunk
+    assert results[0].chunk.id == short_chunk.id
```
KevinJBoyer, line 89: What's the reason for these changes? Because the db_session from the test fixture (the second parameter to test_retrieve_with_scores) is different from the db_session generated by retrieve_with_scores?

KevinJBoyer (approved):
LGTM, just had a question to make sure I understood changes to the test correctly. TY!

app/tests/src/test_retrieve.py
KevinJBoyer: Gotcha. I *think* the identifier here is the literal address in memory: under the hood, SQLAlchemy ensures that if you retrieve the same row back in the same session, it creates only a single instance of the object in memory so you can do things like (psuedocode):
```
with some_session:
    chunk_1 = some_session.select(chunk_id="abc").first()
    chunk_2 = some_session.select(chunk_id="abc").first()
    assert chunk_1 == chunk_2 # these are literally the same object in memory
```

PR #327, opened 2026-03-25
hotfix: pin litellm version between 1.15.2 and 1.82.7

cory-nava:
- [x] Update PR Title to follow this pattern: `[INTENT]: [MESSAGE]`

## Changes
[Hotfix for LiteLLM security vuln](https://www.xda-developers.com/popular-python-library-backdoor-machine/).

## Context for reviewers
The locked version of LiteLLM is `1.63.8`, which is well below the compromised versions (`1.82.7` and `1.82.8`). This project is not affected by this supply chain attack.

However, the [pyproject.toml specifies ^1.51.2](https://github.com/navapbc/labs-decision-support-tool/blob/f2c05b6be9c55bc22c4dc0f899afcf1f2d5827f3/app/pyproject.toml#L24) (compatible with any 1.x >= 1.51.2), so if someone runs `poetry update` or installs without a lock file, they could pull in a compromised version.

I've pinned the version **between** `1.51.2` and `1.82.7`.

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -1,4 +1,4 @@
-# Use the official python3 image based on Debian 11 "Bullseye".
+# Use the official python3 image based on Debian.
 # https://hub.docker.com/_/python
 
 # The build stage that will be used to deploy to the various environments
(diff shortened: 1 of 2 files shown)


MichelleHadfield (approved):

MichelleHadfield (changes requested):
The test failure is in test_url_query_values at line 18: chat_engine.create_engine(engine_id). This call is trying to create an engine for "ca-edd-web", which likely involves connecting to an external service (e.g., LiteLLM or an LLM
  provider) during initialization. The httpx.ConnectError means it's failing to establish that network connection from inside the Docker container.

  This is a network connectivity issue during engine initialization, not a logic bug. The test parses URL query parameters (which works fine), but then calls create_engine() which apparently tries to reach an external HTTP endpoint 
  (likely to validate the LLM connection or fetch model info). That connection is being refused or timing out.
                                                                                                                                                                                                                                        
  Possible causes:                                                                                                                                                                                                                    
  1. LiteLLM or another service that the engine depends on isn't running/accessible from the test container                                                                                                                           
  2. The pinned LiteLLM version (from commit 4ff3d6b on your current branch) may have changed initialization behavior that now makes an HTTP call where it didn't before                                                                
  3. Missing network access in the Docker Compose test environment

cory-nava:
The pinned version of LiteLLM did not break the tests - I believe they've been broken for some time now.

From what Claude said "The Docker image fails to build because Debian removed the apt-key command. I fixed the Dockerfile so it builds again."

Looking at the above, the only failing test is now for deploying the app. Any insights here @MichelleHadfield ?

MichelleHadfield (approved):

